PR #2, opened 2019-05-24
[WIP] Change to relative paths

asantos00:
I had a few problems while working at https://github.com/samselikoff/ember-cli-mirage/pull/1633.

This fixes *some of them*, fixes are not finished (some of them are temporary just to see if I could get it working)

To note:
- We don't have global `require` anymore (at least not now, we have to figure out a way)
- As the files are imported by path on `ember-cli-mirage`, our rollup alias that does the mapping from `@miragejs/server` to the project root does not work (haven't found a solution too)
- `assert` and `require` are *mocked* just so this errors get out of the way when trying to get `ember-cli-mirage` to pass the tests

### DIFF
--- a/lib/orm/associations/association.js
+++ b/lib/orm/associations/association.js
@@ -1,4 +1,4 @@
-import { dasherize } from '@miragejs/server/lib/utils/inflector';
+import { dasherize } from '../../utils/inflector';
 
 /**
   @hide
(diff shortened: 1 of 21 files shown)


samselikoff:
Yep, I think this looks good! I'm actually surprised we still had `@miragejs/server` imports in there, I think that's kinda non-standard even though I absolutely hate that you can't do it...

Looks good for now though 👍 feel free to merge when you're ready

PR #5, opened 2019-05-25
Add inflector customisation api

asantos00:
Adds API to customise `inflected` inflector


samselikoff:
What about wrapping this in a function or property on the server instance?

Want to be able to have multiple server instances simultaneously. 

Even if the instance-based API is modifying the global import, it will still let us improve it in the future without having to change ember-cli-Mirage code. 

What about server.inflector.infections?

asantos00:
I like your suggestion, I'll try to work with instances in inflected too. I think they have support for it. Tying it to server instance should do the job 👍. It is also more coherent with the rest of the API

samselikoff:
Superseded by https://github.com/miragejs/server/pull/6

PR #15, opened 2019-06-23
WIP: Run tests with qunit and testem

asantos00:
First try on running tests standalone

- Adds first version of standalone test running
- Adds temp folder "working-tests" for testing
- Configures testem and qunit (bundling with rollup before running tests)

Still a lot to do, it's a start, but managed to get tests running!!!!!

image


samselikoff:
Very nice man!!

samselikoff:
Can we not share the same rollup config for the tests? I'd worry about the logic drifting between the two (and tests not covering prod build behavior)

samselikoff:
Also I'd say let's keep this PR small and move one test from `test` to `working-test` (deleting it from old place) and get the pattern locked in, then merge it. Then in another PR we can do all the grunt work of copying everything else over.

asantos00:
Regarding the rollup config, we can probably share stuff, not sure if it's going to be possible to share everything (as the test config is to build the tests and so it as a different input). 
However, the code it is running against is the bundled production code

asantos00:
Regarding the test stuff, now these tests are not updated, I would say we have to migrate tests from `ember-cli-mirage` to this repo. What do you meant with "copy one test"? Do you want me to copy one just to work as a test for the "new testing process"?

samselikoff:
I just meant we should get one test working in your `working-tests` directory (e.g. `db-test`), and make sure to delete the original file from `tests` as well, so we can keep track.

We should also get Travis set up to start running tests.

samselikoff:
And yep we can copy from `ember-cli-mirage` to make sure we have the latest versions 👍

asantos00:
Ok, so, should we configure Travis from now? This way we start running tests very soon. I don't think I've access to configure travis

samselikoff:
Should be all set https://travis-ci.org/miragejs/server

samselikoff:
https://github.com/miragejs/server/pull/17 closes

PR #29, opened 2019-07-25
[WIP] Run mirage in node

ryanto:
This PR brings over the previous work from Ember CLI Mirage to run mirage inside of node.

The biggest gotcha here is that pretender is not importable in a node, and doing so throws an error. A good bit of this PR is shimming pretender to an importable class for node (both when testing and building).

The rest of the PR introduces `RequestHandler`, which lets mirage generate a response for requests.

In order to wrap my head around what code is for node and what code is for the browser I created `NodeServer` and `BrowserServer`. They're both very simple.


ryanto:
Ok, here's the next node roadblock:

A bunch of the tests use `fetch` to assert mirage is returning the right values. Since we aren't running an interceptor in node these tests will fail. Options here are use something like nock to match mirage's browser interception behavior.

samselikoff:
Closing for now, we'll keep a reference to it in the Node tracking issue: https://github.com/miragejs/miragejs/issues/210

PR #36, opened 2019-07-29
Use assign method from lodash to make IE work again

stukalin:
The latest ember-cli-mirage stopped working on IE11. After some research I found out that this happens most likely because of the `Object.assign` method (hopefully this is the only reason). Anyways I suggest reverting the changes from the commit 3867bcdb and use the assign method from lodash.

### DIFF
--- a/lib/db-collection.js
+++ b/lib/db-collection.js
@@ -1,10 +1,10 @@
-import { map, isEqual } from "lodash-es";
+import { map, isEqual, assign } from "lodash-es";
 
 function duplicate(data) {
   if (Array.isArray(data)) {
     return data.map(duplicate);
   } else {
-    return Object.assign({}, data);
+    return assign({}, data);
   }
 }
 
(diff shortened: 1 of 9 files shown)


samselikoff:
You're right, and I don't think this is the first time this has happened. We need to come up with a failing test of some sort... even a unit test that says "This project doesn't use Object.assign". Or we need to configure our polyfill rules somehow.

samselikoff:
I wonder if this is better solved by updating [our rollup build](https://github.com/miragejs/server/blob/master/rollup.config.js#L21) to target ie 11

stukalin:
I should give it a try

вт, 30 июл. 2019 г., 16:02 Sam Selikoff <notifications@github.com>:

stukalin:
So I guess the #54 fixes this one. I'll close it then.

samselikoff:
Not 100%, would like to verify

PR #57, opened 2019-08-02
[WIP] Rollup config, and jest integration tests against builds

ryanto:
This needs a check-in / review.

* Changes the rollup config to build for esm, cjs, and umd. `package.json` `module` will point at the esm, and `main` will point at index, which redirects to cjs. 

* Jest will run the integration tests against each of the bundles. These tests will fail, because they're importing files from lib. We can change these to use all public API. Probably should not enable these tests right now.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,7 +9,8 @@ module.exports = {
   ],
   env: {
     "jest/globals": true,
-    es6: true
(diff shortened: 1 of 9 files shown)


ryanto:
dog

samselikoff:
made me lol

PR #105, opened 2019-09-04
Remove assignment to Parent because ugly thinks its a variable that's never used

taras:
When the UMD build is uglified, uglify removes assignment which causes Parent to become undefined. This change prevents uglify from removing the assignment by proactively removing the assignment.

### DIFF
--- a/lib/utils/extend.js
+++ b/lib/utils/extend.js
@@ -4,9 +4,7 @@ import has from "lodash.has";
   @hide
 */
 export default function extend(protoProps, staticProps) {
-  let Parent = this;
-
-  class Child extends Parent {
+  class Child extends this {
     constructor(...args) {
       super(...args);
       // The constructor function for the new subclass is optionally defined by you
@@ -19,7 +17,7 @@ export default function extend(protoProps, staticProps) {
 
   // Add static properties to the constructor function, if supplied.
 
-  Object.assign(Child, Parent, staticProps);
+  Object.assign(Child, this, staticProps);
 
   // Add prototype properties (instance properties) to the subclass,
   // if supplied.


cowboyd:
For some clarification, when babel transpires this down in `mirage.umd.js` you get this:

```js
  function extend(protoProps, staticProps) {
    var Parent = this;

    var Child =
    /*#__PURE__*/
    function (_Parent) {
      _inherits(Child, _Parent);

      function Child() {
        var _getPrototypeOf2;

        var _this;

        _classCallCheck(this, Child);

        for (var _len = arguments.length, args = new Array(_len), _key = 0; _key < _len; _key++) {
          args[_key] = arguments[_key];
        }

        _this = _possibleConstructorReturn(this, (_getPrototypeOf2 = _getPrototypeOf(Child)).call.apply(_getPrototypeOf2, [this].concat(args))); // The constructor function for the new subclass is optionally defined by you
        // in your `extend` definition

        if (protoProps && lodash_has(protoProps, "constructor")) {
          var _protoProps$construct;

          (_protoProps$construct = protoProps.constructor).call.apply(_protoProps$construct, [_assertThisInitialized(_this)].concat(args));
        }

        return _this;
      }

      return Child;
    }(Parent); // Add static properties to the constructor function, if supplied.
```

For some reason, uglify optimizes out the call to `Parent` from the immediately invoked function that babel produces, and then says "oh, well, I guess `Parent` is never used, so let's optimize that out too"

🤷‍♂

ryanto:
👍

samselikoff:
Bizarre. Thanks guys! Psyched to see y'all using this 😄

cowboyd:
Yeah, it's been fun, and can't wait to share more:

Miragejs + Production deploy of brand new app = `mind blown!`

When do you think we can have that version on NPM?

samselikoff:
Let me publish now :)

samselikoff:
https://travis-ci.org/miragejs/server/builds/580899176 should publish when finished!

PR #107, opened 2019-09-05
[WIP] Start testing all builds

ryanto:
Very much a work in progress but I wanted to push so we can discuss.

A few questions came up while doing this:

1) Patching pretender to no-op in umd makes sense, since umd is the entire server with no external dependencies. Also makes sense for cjs+node, since pretender doesn't work there we're forced to patch it because otherwise the consuming app crashes. What about cjs+browser, or esm? These environments might want to import their own pretender, or have a resolution to a specific pretender version. 

2) I am not sure how to rewrite pretender in esm. Hit a wall.

I could see us spending a lot of time trying to answer these questions, so it might be best to setup testing scenarios that reflect how specific tooling consumes these builds.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,6 +1,5 @@
 let internal = {
   displayName: "internal",
-  transformIgnorePatterns: ["<rootDir>/node_modules/(?!lodash)"],
   setupFilesAfterEnv: ["jest-extended"],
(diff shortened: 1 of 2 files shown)


samselikoff:
Thank you for your service. 🇺🇸

Issue #63, opened 2019-08-08
Verify ie11


samselikoff:
If we're committing support for IE11 it'd be nice to have some test coverage for it, somehow

https://github.com/miragejs/server/pull/36

ryanto:
Where did we end up this? Was it that the host app would be responsible for babel+ie11 support?

PR #114, opened 2019-09-06
Rollback polyfills for ie11

stukalin:
Addresses #63

It looks like the commit dda5f85637fe41d6202c894ce377c5bacf87c324 has rolled back the changes were made in the #54 so that the latest version of the server lacks the support of IE11 again (that was close). I've just checked locally that bringing those polyfills back fixes things for ie 11.


samselikoff:
@stukalin oh man. Apologies! We are actually working over in #107 to get tests in place that cover our builds so this doesn't happen again.

Sorry about that, will address this soon 👍

samselikoff:
I *believe* the latest way to do this is with the `useBuiltIns` option to `@babel/preset-env`, which [we are doing here for our umd build](https://github.com/miragejs/server/blob/master/rollup.config.js#L95-L98). The problem is we fixed a typo to our esm build, so I'm guessing your setup started pulling it in, but [we didn't make the change to the babel plugin over there](https://github.com/miragejs/server/blob/master/rollup.config.js#L20) after removing the import statement version.

Let me try that change and we'll see if it works for you. (I don't want to block this on test coverage because it's quite hard and could take longer.)

samselikoff:
I think we'll need to publish to npm so you can try it out due to the build. I'll publish a beta and let you know when it's ready.

samselikoff:
@stukalin could you give `v0.1.24-beta.1` a try?

stukalin:
I will try it in 8 days, for my vacation has just started.

пт, 6 сент. 2019 г., 17:34 Sam Selikoff <notifications@github.com>:

samselikoff:
Sounds good! Let me know then if you run into any issues.

Issue #128, opened 2019-09-10
Factory performance issues


samselikoff:
Issues with using factories on larger data sets keeps coming up so I wanted to open an issue and consolidate the conversations so we can get down to the root of the issue.

It probably has something to do with some much older code somewhere, possibly in `db`.

Ryan was asking about a way to track which factories were taking longest over in https://github.com/miragejs/server/issues/91 and I hacked something together quickly that should give folks a way to do some measurement:

```js
// mirage/config.js
import { Server } from "@miragejs/server";

let originalCreate = Server.prototype.create;
Server.prototype.create = function() {
  console.log("beforeCreate");
  let model = originalCreate.apply(this, arguments);
  console.log("afterCreate");

  return model;
};

export default function() {
  this.get("/users", ...)
}
```

It would be great to get some small demo projects going so we can start to fix this long-standing issue!

I want to help point anyone in the right direction who can help as every time this comes up something else is always more important.

makepanic:
If anyone wants to use it with the performance measure api:

```js
const markerA = 'Mirage create start';
const markerB = 'Mirage create end';

let originalCreate = Server.prototype.create;
Server.prototype.create = function() {
  performance.mark(markerA);
  let model = originalCreate.apply(this, arguments);
  performance.mark(markerB);
  performance.measure('Mirage create', markerA, markerB);

  return model;
};
```

Open devtools performance tab and record.

makepanic:
After removing faker from the generation it seems like most of the time is spend when collecting names (modelName, toCollectionName, ...)

![20190911-090930](https://user-images.githubusercontent.com/1205444/64678687-5cd3e500-d47a-11e9-81e5-c0708d048259.png)

It seems like we could memoize the inflection as it shouldn't change for a word at runtime, right?

https://github.com/miragejs/server/blob/f4d76762651fd2b6e1df71f1725d395bf34881ef/lib/utils/inflector.js#L12

@samselikoff  ould you be open to memoize `camelize` and maybe some name retrieval methods?

samselikoff:
Would definitely be open to this sort of PR! I really have very little experience here so please, I would love to hear your ideas.

I think the fastest way to keep us moving is to POC the change in a small PR so I can give feedback over anything I have input on. But definitely interested in this and willing to take your lead :)

And yes those calls should not change at runtime

makepanic:
I've started working on this in https://github.com/miragejs/server/pull/130

[...]

PR #130, opened 2019-09-11
refactor: memoize some methods

makepanic:
implements parts of #128

### DIFF
--- a/lib/orm/associations/belongs-to.js
+++ b/lib/orm/associations/belongs-to.js
@@ -15,13 +15,13 @@ const identifierCache = {};
  */
 export default class BelongsTo extends Association {
   get identifier() {
-    if (typeof identifierCache[this.key] === "string")
-      return identifierCache[this.key];
+    if (typeof identifierCache[this.key] !== "string") {
+      const identifier = `${camelize(this.key)}Id`;
 
-    const identifier = `${camelize(this.key)}Id`;
+      identifierCache[this.key] = identifier;
+    }
 
-    identifierCache[this.key] = identifier;
-    return identifier;
+    return identifierCache[this.key];
   }
 
   /**
@@ -41,13 +41,13 @@ export default class BelongsTo extends Association {
    */
   getForeignKey() {
     // we reuse identifierCache because it's the same logic as get identifier
-    if (typeof identifierCache[this.key] === "string")
-      return identifierCache[this.key];
+    if (typeof identifierCache[this.key] !== "string") {
+      const foreignKey = `${camelize(this.key)}Id`;
 
-    const foreignKey = `${camelize(this.key)}Id`;
+      identifierCache[this.key] = foreignKey;
+    }
 
-    identifierCache[this.key] = foreignKey;
-    return foreignKey;
+    return identifierCache[this.key];
   }
 
   /**
--- a/lib/orm/associations/belongs-to.js
+++ b/lib/orm/associations/belongs-to.js
@@ -2,6 +2,8 @@ import Association from "./association";
 import { capitalize, camelize } from "../../utils/inflector";
 import assert from "../../assert";
 
+const identifierCache = {};
+
 /**
  * The belongsTo association adds a fk to the owner of the association
  *
@@ -13,7 +15,13 @@ import assert from "../../assert";
  */
 export default class BelongsTo extends Association {
   get identifier() {
-    return `${camelize(this.key)}Id`;
+    if (typeof identifierCache[this.key] === "string")
+      return identifierCache[this.key];
+
+    const identifier = `${camelize(this.key)}Id`;
+
+    identifierCache[this.key] = identifier;
+    return identifier;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


makepanic:
I've added memoization to the association classes. It might be unsafe but our tests and mirage tests work fine with it.
The issue with these methods is that they don't get an argument. Alternatively we could extract private methods which receive the instance fields as arguments to make them safe to memoize.

makepanic:
I've also ran a slightly larger test, which uses more mirage instances to check how it compares:

* old mirage

1416ms
1290ms
1127ms
1185ms
1172ms
1143ms
1145ms

* memoized mirage

1053ms
969ms
938ms
909ms
973ms
969ms
918ms

Seems like a constant improvement.

chbonser:
I can confirm that I saw a large improvement with this change set. Models which were taking ~300ms to create are now taking ~100ms.

I will note that I did have to modify db-collection.js as follows as sometimes the `id` is `null`. I did not see this with the master branch.

```
@@ -344,7 +344,7 @@ class DbCollection {
     @hide
    */
   _findRecord(id) {
-    id = id.toString();
+    id = id ? id.toString() : id;
```

makepanic:
> I will note that I did have to modify db-collection.js as follows as sometimes the `id` is `null`. I did not see this with the master branch.
> 

That's odd.
Maybe we should memoize ourself and not let lodash do it. There might be some unexpected lodash features that play into it.

makepanic:
@chbonser I've refactored some of the association memoized methods. Now they're native getters that call the methods by passing an argument. Can you check if your adjustment is still needed?

@samselikoff I'm not sure if the latest commit is fine as it's potentially breaking api consumers due to multiple api changes.

chbonser:
@makepanic That did the trick.

samselikoff:
@makepanic Yeah I mean these methods have never been documented, there's no way every method in Mirage is public. I think they are not breaking changes but we should still clearly communicate the changes.

Great job on all this work, super appreciate of you diving into this!

So using a getter that calls a proxied function is faster eh?

makepanic:
> So using a getter that calls a memoized function is faster eh?

I don't think that there's a perf difference. 
It's just that a pure function is easier to memoize if its arguments are passed in. Previously it was memoized while still reading instance values. Repeated calls would've returned the initial "cached" result.

lib/orm/associations/belongs-to.js
```diff
@@ -13,7 +15,13 @@ import assert from "../../assert";
  */
 export default class BelongsTo extends Association {
   get identifier() {
-    return `${camelize(this.key)}Id`;
+    if (typeof identifierCache[this.key] === "string")
```
ryanto: Any reason for the `typeof` check vs `if (identifierCache[this.key])`?

lib/orm/associations/belongs-to.js
```diff
@@ -13,7 +15,13 @@ import assert from "../../assert";
...
+      return identifierCache[this.key];
+
+    const identifier = `${camelize(this.key)}Id`;
+
+    identifierCache[this.key] = identifier;
+    return identifier;
   }
```
ryanto: I'd like to avoid early and double returns in these methods if possible. I'm thinking something like this might be a bit more readable.

```js
get identifier() {
  if (typeof identifierCache[this.key] !== "string") {
    const identifier = `${camelize(
      this._container.inflector.singularize(this.key)
    )}Ids`;

    identifierCache[this.key] = identifier;
  }

  return identifierCache[this.key];
}
```

Is there any downside to setting the cache if the value doesn't exist, and then returning from the cache?

lib/orm/associations/belongs-to.js
makepanic: The return type is always a string and doing a typeof check prevents theoretical false-positives, like `key: ""` which would never use the cache.

I can remove it as an empty string key is probably not likely to appear.

lib/orm/associations/belongs-to.js
makepanic: I don't think that there's a downside. 
I just try to avoid the additional nesting (which is probably cleaner in other languages, see e.g. [kotlin](https://github.com/Turbo87/intellij-emberjs/blob/27d45ad1b50f1f1476b9cd50f07abac77e32bca2/src/main/kotlin/com/emberjs/psi/EmberRelationshipAnnotator.kt#L29))

If you usually do it that way, I'll refactor it to match your style.

lib/orm/associations/belongs-to.js
ryanto: Ok cool, great point. Let's keep the `typeof` then!

lib/orm/associations/belongs-to.js
ryanto: Yup we usually write in this style so I think it makes sense to refactor these.

makepanic: @ryanto I've commited some changes. 
Can you check if I understood your request correctly? I'll squash it after it's ok.

ryanto: Yup, looks great 👍

ryanto:
🎉 🎉 🎉

Nice job @makepanic

PR #132, opened 2019-09-12
refactor: replace some filter()[0] usages to avoid unneeded iterations

makepanic:
This probably won't affect performance in the common case.
This removes just some unnecessary array method calls.

### DIFF
--- a/lib/orm/model.js
+++ b/lib/orm/model.js
@@ -714,9 +714,7 @@ class Model {
     if (Array.isArray(foreignKeys)) {
       let association = Object.keys(this.hasManyAssociations)
         .map(key => this.hasManyAssociations[key])
-        .filter(
-          association => association.getForeignKey() === foreignKeyName
-        )[0];
+        .find(association => association.getForeignKey() === foreignKeyName);
 
       let found;
       if (association.isPolymorphic) {
@@ -742,9 +740,7 @@ class Model {
     } else {
       let association = Object.keys(this.belongsToAssociations)
         .map(key => this.belongsToAssociations[key])
-        .filter(
-          association => association.getForeignKey() === foreignKeyName
-        )[0];
+        .find(association => association.getForeignKey() === foreignKeyName);
 
       let found;
       if (association.isPolymorphic) {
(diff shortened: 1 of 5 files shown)


lib/orm/model.js
```diff
@@ -824,15 +824,18 @@ class Model {
...
+            // filter out models that will still be associated
+            !tempAssociation.includes(associate) &&
+            associate.hasInverseFor(association)
+        )
         .forEach(associate => {
-          if (associate.hasInverseFor(association)) {
             let inverse = associate.inverseFor(association);
 
             associate.disassociate(this, inverse);
             associate.save();
```
jelhan: Indentation must be adjusted after `if` block is removed.

lib/orm/polymorphic-collection.js
```diff
@@ -118,7 +118,7 @@ export default class PolymorphicCollection {
    * @public
    */
   includes(model) {
-    return this.models.filter(m => isEqual(m.attrs, model.attrs)).length > 0;
+    return this.models.find(m => isEqual(m.attrs, model.attrs)) !== undefined;
```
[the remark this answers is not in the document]
makepanic: Changed it to some as it makes it easier to understand.

lib/orm/model.js
makepanic: You're right. I didn't see that prettier is usable in this project. I ran prettier on the changed files and rebase + pushed the changes.

samselikoff: thanks for the review @jelhan 😄 Very helpful to have other folks checking this out!

@makepanic I think I did install prettier here but still need to (a) go through and prettierify everything and (b) figure out the best way to enforce this (husky? tests?)

makepanic: We're using eslint with https://github.com/prettier/eslint-plugin-prettier which fails the linting stage if prettier has issues with our code.

makepanic: For running prettier on all project files, you could add a package.json script like:
```
    "prettier": "prettier --write \"{config,app,tests}/**/*.{js,ts}\"",
```
And adjust it to the mirage source files.

makepanic:

ryanto:
Thanks @makepanic !

PR #134, opened 2019-09-15
refactor: reduce loop count during model setup

makepanic:
This reduces the amount of loops that are done during hydration when setting models up.

See example on how often `_setupAttrs` and `_setupRelationships` are called:

This also replaces some `arr.indexOf()` calls with `arr.includes` as includes is already used elsewhere to check if an array includes an item).

**Note**: this has to rebased if the other PRs are merged.

I could also move all the changes into one bigger PR if that's easier to review.

### DIFF
--- a/lib/orm/model.js
+++ b/lib/orm/model.js
@@ -81,7 +81,7 @@ class Model {
       @property attrs
       @public
     */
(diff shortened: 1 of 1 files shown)


makepanic:

PR #151, opened 2019-10-01
perf: use Set for association(Id)Keys

makepanic:
Both `associationKeys` and `associationIdKeys` are primarily used to check if a key is included in them.
They're looked up repeatedly when setting up attrs.

I've changed them to a Set because:
* `set.has` should be something like closes to O(1)  (see also [microbenchmark](http://jsben.ch/TFCDD)`*`)
* Set is still [supported by IE11](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#Browser_compatibility)

This PR also extracts the static arguments from server.create/createList as they never change on each invocation.

`*` = microbenchmarks are never perfect

Using https://gist.github.com/makepanic/241013f4dd94d6064d046642d60c43d2 to benchmark this change, I'm getting:

// master
setup x 13.14 ops/sec ±16.22% (24 runs sampled)
setup x 12.84 ops/sec ±17.21% (24 runs sampled)

// use-set
setup x 13.29 ops/sec ±17.00% (24 runs sampled)
setup x 13.22 ops/sec ±16.53% (24 runs sampled)

This doesn't seem to be a big improvement though. It's up to you if you want it.

### DIFF
--- a/lib/orm/associations/belongs-to.js
+++ b/lib/orm/associations/belongs-to.js
@@ -76,8 +76,8 @@ export default class BelongsTo extends Association {
 
     // TODO: look how this is used. Are these necessary, seems like they could be gotten from the above?
     // Or we could use a single data structure to store this information?
-    modelPrototype.associationKeys.push(key);
-    modelPrototype.associationIdKeys.push(foreignKey);
+    modelPrototype.associationKeys.add(key);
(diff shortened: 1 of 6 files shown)


samselikoff:
Awesome, Set seems good. I'd love to have automated ie11 coverage bc I wouldn't want to regress but it does seem to be supported.

PR #152, opened 2019-10-02
perf: make validation of foreign keys a lookup instead of iterating over all of them

makepanic:


makepanic:
This is a small change to improve _validateForeignKeyExistsInDatabase performance.

`_validateForeignKeyExistsInDatabase` is called for each foreign key attribute when creating any model.

baseline:


this change:


(Both trees are on https://github.com/miragejs/server/pull/151 in one of our tests)

Couple thinks that could be discussed:
* should updating of the fk -> association tables be extracted into methods?

ryanto:
Hey @makepanic this looks great! Looks like it needs to be rebased.

makepanic:
rebased on latest master

PR #153, opened 2019-10-03
perf(model): remove attr iterations when creating a model instance

makepanic:
This removes the times mirage iterates over model attributes during setup.

Before it iterated over n attrs:

* 2x in `_setupAttrs`
* 1x in `_setupRelationships`

This doesn't seem that much but a model instance is created all the time due to [hydration](https://github.com/miragejs/server/blob/master/lib/orm/schema.js#L537).

This PR changes the behavior to only iterate once over all attrs.

This will also allow further optimization to skip relationship setup if the attr is an attribute (I'm not sure if mirage allows that though).

Using https://gist.github.com/makepanic/241013f4dd94d6064d046642d60c43d2 to benchmark this change, I'm getting:

// master
setup x 13.14 ops/sec ±16.22% (24 runs sampled)
setup x 12.84 ops/sec ±17.21% (24 runs sampled)

// model-cycles
setup x 13.32 ops/sec ±16.96% (24 runs sampled)
setup x 13.45 ops/sec ±17.09% (25 runs sampled)

(as usual benchmarks might be misleading)

### DIFF
--- a/lib/orm/model.js
+++ b/lib/orm/model.js
@@ -85,8 +85,18 @@ class Model {
 
     attrs = attrs || {};
 
-    this._setupAttrs(attrs);
-    this._setupRelationships(attrs);
+    // Ensure fks are there
+    this.fks.forEach(fk => {
+      this.attrs[fk] = attrs[fk] !== undefined ? attrs[fk] : null;
+    });
+
+    Object.keys(attrs).forEach(name => {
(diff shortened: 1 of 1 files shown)


ryanto:
Looks great, ready to merge! Looks like it needs a rebase.

makepanic:
rebased on current master

PR #166, opened 2019-10-15
route-handlers/shorthands: Return 404 from PUT/DELETE if model can't be found

Turbo87:
Before this PR the shorthand route handlers would return a Internal Server Error because `.find(id)` returns `undefined` and calling a method on `undefined` throws an error. This PR adds a check to both handlers that matches the behavior of the `GET` shorthand handler.


samselikoff:
Nice - thanks!

PR #181, opened 2019-10-28
fix: fix trailing slash regression during url model extraction

### LINKED ISSUE #180: Shorthand route fails if it has a trailing slash
In our app we had a shorthand route set up as `this.get('/notifications/');` which broke our tests when upgrading from ember-cli-mirage 1.1.3 to 1.1.4 with the following error:

`"Mirage: The route handler for http://localhost:4200/ghost/api/v3/admin/notifications/ is trying to access the  model, but that model doesn't exist."`

Digging in, it appears that `getModelClassFromPath()` is failing when the path has a trailing slash:

Screenshot 2019-10-28 at 15 35 27

makepanic:
fixes #180


makepanic:
seems like the failure is unrelated to the change

kevinansfield:
Thanks @makepanic, I tested this and it's fixed our test suite 👍

makepanic:

samselikoff:
Awesome thanks for the quick fix @makepanic !

PR #187, opened 2019-11-04
Throw when association helper is used with polymorphic belongsTo

krasnoukhov:
I stumbled upon this while working on the app so decided why not fix it :-)

Basically foreign key validation for polymorphic expects `{ type, id }` object here:

https://github.com/miragejs/miragejs/blob/4ab442d1570625e8ccfdfed04f5c7c87d1269350/lib/orm/model.js#L712-L714

### DIFF
--- a/__tests__/external/shared/factories/helpers-test.js
+++ b/__tests__/external/shared/factories/helpers-test.js
@@ -136,4 +136,27 @@ describe("Eternal | Shared | Factories | helpers", () => {
       server.create("page");
     }).toThrow();
   });
+
+  test("it allows to use association helper with polymorphic relationship", () => {
+    server = new Server({
+      environment: "test",
+      models: {
+        author: Model.extend({
+          anyPost: belongsTo("base-post", { polymorphic: true })
+        }),
+        basePost: Model.extend({})
+      },
+      factories: {
+        author: Factory.extend({
+          anyPost: association()
+        }),
+        basePost: Factory.extend({
+          title: "base title"
+        })
+      }
+    });
+
+    let author = server.create("author");
+    expect(author.anyPost.title).toEqual("base title");
+  });
 });
--- a/__tests__/external/shared/factories/helpers-test.js
+++ b/__tests__/external/shared/factories/helpers-test.js
@@ -137,7 +137,7 @@ describe("Eternal | Shared | Factories | helpers", () => {
     }).toThrow();
   });
 
-  test("it allows to use association helper with polymorphic relationship", () => {
+  test("it throws if using the association helper with polymorphic relationship", () => {
     server = new Server({
       environment: "test",
       models: {
@@ -149,14 +149,12 @@ describe("Eternal | Shared | Factories | helpers", () => {
       factories: {
         author: Factory.extend({
           anyPost: association()
-        }),
-        basePost: Factory.extend({
-          title: "base title"
         })
       }
     });
 
-    let author = server.create("author");
-    expect(author.anyPost.title).toEqual("base title");
+    expect(() => {
+      server.create("author");
+    }).toThrow();
   });
 });
--- a/lib/server.js
+++ b/lib/server.js
@@ -1266,22 +1266,21 @@ export default class Server {
           `You're using the association() helper on your ${modelName} factory for ${attr}, which is a belongsTo self-referential relationship. You can't do this as it will lead to infinite recursion. You can move the helper inside of a trait and use it selectively.`
         );
 
+        let isPolymorphic =
+          association && association.opts && association.opts.polymorphic;
+
+        assert(
+          !isPolymorphic,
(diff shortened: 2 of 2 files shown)


krasnoukhov:
While this fixes the crash, it would also be nice to be able to provide the actual relationship model name for the factory-based creation since base polymorphic type might not be a good default in some cases. @samselikoff do you have any thoughts about that?

ryanto:
Thanks!

Two quick things...

1. I think when defining polymorphic relationships we don't specify a model name, this is so any model can be put into that type.
1. I would expect the `association` factory helper to not work with polymorphic relationships because it doesn't know which model to use for the relationship.

```js
new Server({
  models: {
    post: Model.extend({
      comments: hasMany()
    }),
    picture: Model.extend({
      comments: hasMany()
    }),
    comment: Model.extend({
      commentable: belongsTo({ polymorphic: true })
    },
  }
  factories: {
      post: Factory.extend({}),
      picture: Factory.extend({}),
      comment: Factory.extend({
        commentable: association() // how do we know what to use here, could be post or picture?
      })
  }
})
```

The way we recommend solving this is using `afterCreate` to assign the polymorphic relationship.

```js
comment: Factory.extend({
  afterCreate(comment, server) {
    comment.update({
      commentable: server.create('post')
    })
  }
})
```

You can also use multiple traits to create different polymorphic associations. A `withPost` or `withPicture` trait on the comment factory would do this.

Let me know if I'm understanding this correctly, if you have more use cases feel free to add 'em!

Thanks again!

krasnoukhov:
@ryanto Thanks for the quick reply! What you say totally makes sense, I should've also considered polymorphic with no model name.

I can definitely use `afterCreate` but it's pretty awkward as opposed to declarative `association`, specifically in a case when I want to have relationship created by default in the factory as well as in the `trait`, but at the same time supporting manually passed value. In the trait, I can't really use the `if (!comment.commentable)` because it would always be there already because of the default factory, and I can't omit the check because I don't want to override whatever was passed manually. Does this make sense?

Just had this thought, what if we extend `association` API to support model name for the polymorphic use case?

```js
comment: Factory.extend({
  commentable: association().modelName('post'),

  forPicture: trait({
    commentable: association().modelName('picture'),
  }),
})
```

Specifying the model name would be required when used with polymorphic, otherwise helpful error message will be shown. This way everything should work smoothly with default factory, trait and manual values. Also does not seem like this would be a big change :-)

If that's not something you'd be wiling to accept that's fine I guess, just let me know and in that case I'll repurpose this PR to rather throw an error when `association` is used with polymorphic relationship so future users will be less confused than me.

ryanto:
Yup, absolutely extending the `association()` helper to take some configuration for these situations would be helpful.

I think at this point since there's a way around it with `afterCreate`, although it's not ideal and clunky, is the best path forward. At this point in time we don't want to add more API surface area to the project because we have higher priorities like getting the website finished and adding node support. I think we'll be happy to re-visit the factory APIs in the future when we have more bandwidth. 

Let me know your thoughts on this.

krasnoukhov:
@ryanto Ok, I added an assertion here that relationship is not polymorphic instead.

Another API idea I had would be to make association take a closure:

```js
comment: Factory.extend({
  commentable: association((server) => server.create('post')),

  forPicture: trait({
    commentable: association((server) => server.create('picture')),
  }),
})
```

Maybe that'd be even more helpful in other non-polymorphic cases as well.

Just to be clear, I'd totally be wiling to implement the API change here cause this is something we'd definitely utilize across our apps. So please hit me up when you're ready to address this problem, thanks :-)

samselikoff:
@krasnoukhov 👏 nice work, thank you very much for the test sir!

We have some other changes we want to make to the Factory API (including accepting params from the calling site & migrating to true ES6 classes), so I think I want to hold off on adding a feature like this until potentially later.

For now you should be able to just use `afterCreate` to write your factory like this:

```js
comment: Factory.extend({
  afterCreate(comment, server) {
    comment.update({ commentable: server.create('post') })
  },

  forPicture: trait({
    afterCreate(comment, server) {
      comment.update({ commentable: server.create('picture') })
    }
  }),
})
```

krasnoukhov:
My pleasure @samselikoff! Thanks for the suggestion, we'll definitely be using `afterCreate` for now. 

Please check out this case when considering better API for `association`, this is currently the major source of awkwardness with `afterCreate`:

> I can definitely use `afterCreate` but it's pretty awkward as opposed to declarative `association`, specifically in a case when I want to have relationship created by default in the factory as well as in the `trait`, but at the same time supporting manually passed value. In the trait, I can't really use the `if (!comment.commentable)` because it would always be there already because of the default factory, and I can't omit the check because I don't want to override whatever was passed manually. Does this make sense?

samselikoff:
Ah, oh wow – totally missed this, and I'm a little surprised about it!

Going to open a new issue to discuss :)

PR #196, opened 2019-11-07
Update documentation for `schema.js`

corrspt:
I had an issue with findBy, so I updated the docs a little bit.

### DIFF
--- a/lib/orm/schema.js
+++ b/lib/orm/schema.js
@@ -256,18 +256,14 @@ export default class Schema {
 
   /**
     Returns the first model in the database that matches the key-value pairs in `attrs`. Note that a string comparison is used.
-    Also remeber that attribute names are dasherized (for attributes with more than 1 word, like "createdAt" becomes "created-at")
 
     ```js
     let post = blogPosts.findBy({ published: true });
+    let post = blogPosts.findBy({ authorId: 1, published: false });
+    let post = blogPosts.findBy({ author: janeSmith, featured: true });
     ```
-    
-    Remember that relantionships are represented with a camelized `relationNameid` attribute name, so if you want to search for an author, you would do:
-    ```js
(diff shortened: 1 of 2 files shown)


lib/orm/schema.js
```diff
@@ -256,10 +256,16 @@ export default class Schema {
 
   /**
     Returns the first model in the database that matches the key-value pairs in `attrs`. Note that a string comparison is used.
+    Also remeber that attribute names are dasherized (for attributes with more than 1 word, like "createdAt" becomes "created-at")
```
samselikoff: Hm – I don't think this is true. I think maybe somewhere else a dasherized compound word must have been used.

Could you try with a new app that uses a model with a compound attribute name, and see if it's the case?

corrspt: Hi Sam,

I'll do that. Might take me a little while to setup and test that, but I'll get back when I do!

asantos00: Minor: Typo in `remember`

lib/orm/schema.js
```diff
@@ -256,10 +256,16 @@ export default class Schema {
...
 
     ```js
     let post = blogPosts.findBy({ published: true });
     ```
+    
+    Remember that relantionships are represented with a camelized `relationNameid` attribute name, so if you want to search for an author, you would do:
```
asantos00: typo: `relationships`

lib/orm/schema.js
corrspt: Thanks Alexandre for the typo report

samselikoff:
Simplified this a bit by just adding some more examples. Thanks @corrspt!

corrspt:
Thanks so much for moving this forward @samselikoff. I wanted to contribute, but couldn't find the time again.

Issue #184, opened 2019-10-30
JSONAPISerializer should handle empty include string


samselikoff:
e.g. `GET /users/1?include=`

Mirage currently throws, we should just handle this case.

cadeParade:
It also throws if there is a trailing comma in the include param ex: `GET /users/1?include=address,`

samselikoff:
https://github.com/miragejs/miragejs/pull/211 closes

PR #211, opened 2019-11-20
Bug/#184

danieltolmon:
solved issue #184 
JSONAPISerializer should handle empty include string: was test and it worked. 
JSONAPISerializer should handle a trailing comma:  was solved filtering the array of includes and tested.


ryanto:
Looks good! 👍

__tests__/external/shared/serializers/json-api-serializer/associations/includes-test.js
```diff
@@ -519,4 +519,70 @@ describe("External | Shared | Serializers | JSON API Serializer | Associations |
...
+    expect(result).toEqual({
+      data: {
+        type: "blog-posts",
+        id: "1",
+        attributes: {}
+      }
+    });
+  });
+
+  test("Do not throw when queryParamIncludes has a trailing comma", () => {
```
samselikoff: Could we move these tests to the higher level route handler test below? I don't want us to miss something that we might not be catching it here since we're calling `serialize()` method directly instead of sending a request like `/posts/1?include=comments,` to the server.

I think two tests for this PR would suffice, one you basically have already down there. Just can move the trailing comma one there too. And I think add assertion against the data returned.

samselikoff:
@danieltolmon Awesome work + thanks so much for jumping in on this! Just left one comment, after that I think this is good to go.

danieltolmon:
Your welcome! I just published a test in the route handlers to ensure it reads the url when there is trailing commas.

__tests__/internal/move-after-handle-request/route-handlers/function-handler/basic-test.js
```diff
@@ -123,4 +136,47 @@ describe("Integration | Route handlers | Function handler", () => {
...
+
+    server.get("/posts/:id", (schema, request) => {
+      return schema.posts.find(request.params.id);
+    });
+
+    let res = await fetch("/posts/1?include=author,");
+    expect(res.status).toBe(200);
+    expect(JSON.parse(res._bodyInit).included).toEqual([
+      { type: "authors", id: "1", attributes: { name: "Daniel" } }
+    ]);
```
samselikoff: Can we follow the pattern from the other tests, and use `let data = await res.json()` here (`res._bodyInit` looks like a private method so we want to avoid that) and also assert against the shape of the data? Just enough to make sure the right data is there.

danieltolmon: Sorry for the delay, I was in other projects and Christmas came... Happy new year!!
Well, regarding your feedback, I took away the lower case test and I added the data expectation to the high level one.
In the high level test, I did also follow the pattern  from the other tests.
I hope it's all good! 
Cheers

samselikoff:
Thanks for still working on this!

My apologies – looking at the changes, I think moving the tests to function handler test makes it more confusing. We shouldn't be mixing ActiveModelSerializer and JSONAPISerializers like this in a test if we can avoid it, because I think it won't be clear to future readers why they're there.

Currently your code has 1 test in in the `includes-test` module (for checking if includes is empty), and 2 in the function handler test (one for if includes is empty and another for a trailing comma).

Can we remove the second 2, and bring back the trailing comma test to the `includes-test` module? I believe you deleted it in [this commit](https://github.com/miragejs/miragejs/pull/211/commits/33f78df21abda1a578440cd9c6dc9d5c3887e18a). So we'll end up with just 2 tests and your `trim` changes to the JSON API serializer.

(Sorry about nit picking, we've just been bitten by tests in the past that make it hard for us to change/refactor so I only want the absolute necessary in here!)

danieltolmon:
yeap, no worries, its an easy change... and I understand, testing is very important to be organised.

But, just to be sure.... what do you mean with 

> Can we remove the second 2

So I delete both tests (empty and trailing coma) in the function-handler? and we just have 2 tests (empty and trailing coma) in the `includes-test`?

samselikoff:
Correct! Since they're really about the particulars of JSONAPISerializer.

__tests__/internal/move-after-handle-request/route-handlers/function-handler/basic-test.js
```diff
@@ -7,13 +14,19 @@ describe("Integration | Route handlers | Function handler", () => {
     server = new Server({
       environment: "development",
       models: {
-        user: Model.extend({})
+        user: Model.extend({}),
+        post: Model.extend({
+          author: belongsTo()
+        }),
+        author: Model
```
samselikoff: Could you get rid of all the changes to this file since we're not using it anymore?

samselikoff:
Just a little comment and then we can merge this! Sorry for letting this languish.

Issue #185, opened 2019-10-30
Better logging when Mirage route handler throws error


samselikoff:
If a route handler throws an error it can be hard to see from the console which request made the error. We should improve this.

samselikoff:
To elaborate:

There's a few different ways Mirage can "error".

We have an an [assert module](https://github.com/miragejs/miragejs/blob/master/lib/assert.js#L22) that we use to throw some useful validation errors. You can see from those modules that we commented out the console.error lines. I think it was because they were causing noise in testing.

The other way Mirage can error is if a user writes a buggy route handler:

```js
this.get('/movies', () => {
  return foo.bar; // exception: calling bar on undefined object foo
})
```

The way this shows up in different environments can be crappy/hidden.

What I want to do (or for someone to do!) is to go through these situations in different environments: Create React App, new Vue CLI app, probably those two to start. And (1) write a Mirage route handler that fails a Mirage assertion, and (2) write a buggy function handler. Do those in development, and let's see what the resulting error is & how it appears in the console.

danieltolmon:
Ok, I am going to work in this today!

danieltolmon:
Hi @samselikoff ,

This is a screenshot from the console in CRA: 
![image](https://user-images.githubusercontent.com/44772295/69357283-bb88bc80-0c7c-11ea-838e-f5235f9eb170.png)

And the server that creates the error:
![image](https://user-images.githubusercontent.com/44772295/69357549-2cc86f80-0c7d-11ea-8a93-99bd16d13677.png)

I think the message clarifies quite good what's the problem. What do you think? would you like to change something?

samselikoff:
Thanks for this, exactly what I wanted to see!

So, the message is clear, but the problem is if you don't happen to expand the Mirage line. You will have no idea that anything happened.

I think if your Mirage server throws an exception there should also be a `console.error()` printed to the screen.

danieltolmon:
Ok, I think it is a good idea. I can add a console.error() with the message to be printed.

danieltolmon:
Hi @samselikoff, I added a console.error() just when there is an error in the route handler. I also hide it and check it in the testings. I will do a PR, see what you think.

This is how it appears now in the console
![Screenshot 2019-11-25 at 11 22 20](https://user-images.githubusercontent.com/44772295/69536598-26384180-0f76-11ea-8a42-f15ffefbfd26.png)

Cheers!

samselikoff:
[...]

PR #214, opened 2019-11-21
added MirageError when handling errors in route-handler

danieltolmon:
Solved #185 
I added a MirageError when catching an error on the routerHandler. So the user will have more information about the error.
Is it that you where thinking when you reported the issue?

### DIFF
--- a/__tests__/internal/move-after-handle-request/route-handlers/function-handler/basic-test.js
+++ b/__tests__/internal/move-after-handle-request/route-handlers/function-handler/basic-test.js
@@ -32,7 +32,7 @@ describe("Integration | Route handlers | Function handler", () => {
     let res = await fetch("/users");
     let data = await res.json();
 
-    expect(data.message).toBe("I goofed");
+    expect(data.message).toBe("Mirage: I goofed");
     expect(data.stack).toMatch(
       "Mirage: Your GET handler for the url /users threw an error"
     );
--- a/lib/route-handler.js
+++ b/lib/route-handler.js
@@ -95,12 +95,12 @@ export default class RouteHandler {
         result = new Response(
           500,
           {},
-          {
+          new MirageError(
             message,
-            stack: `Mirage: Your ${request.method} handler for the url ${
+            `Mirage: Your ${request.method} handler for the url ${
               request.url
             } threw an error:\n\n${e.stack || e}`
-          }
+          )
         );
       }
     }


samselikoff:
@danieltolmon I'd like to take a broader look at how Mirage reports errors in different environments. I think there are some common situations where the errors get swallowed, and it's hard for the user to know what happened.

Maybe we can start with a single environment – Create React App. Start a new CRA, add Mirage, and then take some screenshots of the console when Mirage errors. We can bring the screens back to #185 to discuss.

What do you think?

danieltolmon:
Yeap @samselikoff , I can do that. and we discuss it on the issue chat

__tests__/internal/move-after-handle-request/route-handlers/function-handler/basic-test.js
```diff
@@ -29,10 +29,13 @@ describe("Integration | Route handlers | Function handler", () => {
       throw "I goofed";
     });
 
+    let consoleSpy = jest.spyOn(console, "error").mockImplementation(() => {});
     let res = await fetch("/users");
     let data = await res.json();
 
-    expect(data.message).toBe("I goofed");
+    expect(consoleSpy).toHaveBeenCalled();
```
asantos00: What do you think of using `toHaveBeenCalledWith` to assert that it is being called with the right message?

danieltolmon: Good Idea! Changed it!

samselikoff:
Great start!

Checking this out now. So, when I drop a simple error in a route handler

```js
this.get("/api/call", () => {
  return window.foo.bar;
});
```

this is what I see:

image

I think it's still confusing for new Mirage users, because there's no connection to the error + their Mirage code. Only if they happen to dig in a bit more to the Mirage response will they see:

image

but this is quite confusing.

If I change it to something that ESLint recognizes as an error

```js
this.get("/api/call", () => {
  return foo.bar;
});
```

this is what I see in the console:

image

I think if our error messages looked like this it would be much more clear. Now, the line number stuff is tricky for us, but if you look at the messaging we do have some information:

> "Mirage: Mirage: Your GET handler for the url /api/call threw an error:

> ReferenceError: foo is not defined

I think if the error showed this, that would be the most helpful thing. So let's prepend Mirage to the error message, and have it say:

```
[ Mirage ] Your GET handler for the url /api/call threw an error:

  foo is not defined
```

including the formatting + new lines (indent the error and prepend message with `[Mirage]`.

Let me know if you have any questions!

danieltolmon:
Okay, I changed the console.error to give more information. Here I send a screenshot about how it looks:

Screenshot 2020-01-09 at 12 07 39

What do you think?

Issue #225, opened 2019-11-28
Add Typescript type files


ZengLawrence:
Please make Typescript files available to make it easier to code in Typescript.

nemonemi:
This ticket is 2 years old, and the project hasn't seen any new releases since the third quarter of 2020.

What has been happening with Typescript in this project?

martis-git:
Still actual!

IanVS:
I believe this should have been closed by #262.

PR #262, opened 2020-01-03
Add TypeScript support to Mirage.js #225

zoltan-nz:
https://github.com/miragejs/miragejs/issues/225

An initial draft to add TypeScript support to Server class.

Any feedback is welcome. If you have any question or suggestion please don't hesitate to comment here.


types/index.d.ts
```diff
@@ -0,0 +1,38 @@
...
+  public createList(type: any, amount: any, traitsAndOverrides?: any): any;
+  public loadFixtures(...args: string[]): any;
+  public passthrough(...paths: string[]): any;
+  public passthrough(fn: (request: any) => any): any;
+  public shutdown(): any;
+
+  public get(path: string, model: string | (() => any)): any;
+  public post(path: string, model: string | (() => any)): any;
+  public put(path: string, model: string | (() => any)): any;
+  public del(path: string): any;
```
asantos00: These method do not return anything, I suggest changing the last `any` to `void`

types/index.d.ts
```diff
@@ -0,0 +1,38 @@
...
+  routes?: (this: Server) => any;
+  baseConfig?: (this: Server) => any;
+  testConfig?: (this: Server) => any;
+}
+
+declare class Server {
+  public db: any;
+  public inflector: any;
+  public logging: any;
+  public namespace: any;
```
asantos00: `namespace` should be a `string`

types/index.d.ts
```diff
@@ -0,0 +1,38 @@
...
+  public pretender: PretenderServer;
+  public schema: any;
+  public timing: any;
```
asantos00: `timing` is a `number`

types/index.d.ts
```diff
@@ -0,0 +1,38 @@
...
+  public urlPrefix: any;
```
asantos00: `urlPrefix` is a `string`

types/index.d.ts
```diff
@@ -0,0 +1,38 @@
...
+  environment?: string;
+  routes?: (this: Server) => any;
+  baseConfig?: (this: Server) => any;
+  testConfig?: (this: Server) => any;
+}
+
+declare class Server {
+  public db: any;
+  public inflector: any;
+  public logging: any;
```
asantos00: `logging` is a `boolean`

types/index.d.ts
```diff
@@ -0,0 +1,38 @@
...
+  constructor(serverOptions: IServerOptions);
+
+  public testConfig(): any;
+  public baseConfig(): any;
+
+  public create(type: any, traitsAndOverrides?: any): any;
+  public createList(type: any, amount: any, traitsAndOverrides?: any): any;
+  public loadFixtures(...args: string[]): any;
+  public passthrough(...paths: string[]): any;
+  public passthrough(fn: (request: any) => any): any;
```
asantos00: This return type is also a `void`

.eslintignore
```diff
@@ -4,3 +4,6 @@
 
 # misc
 /coverage/
+
+#types
```
asantos00: minor: a space after `#`

zoltan-nz:
@asantos00 Thanks for your suggestions, really appreciate. I just implemented those. You right, `schema` and `db` would be a great addition... It would be also great if we can use Mirage in Angular applications. ;)

types/index.d.ts
```diff
@@ -1,19 +1,20 @@
-import { Server as PretenderServer } from "pretender";
+import { RequestHandler, Server as PretenderServer } from "pretender";
 
 declare interface IServerOptions {
   environment?: string;
   routes?: (this: Server) => any;
   baseConfig?: (this: Server) => any;
   testConfig?: (this: Server) => any;
+  seeds?: (server: Server) => any;
```
zoltan-nz: I haven't seen `seeds` in the API doc, however it is mentioned in the Guide.

API: https://miragejs.com/api/classes/server
Guide: https://miragejs.com/docs/data-layer/database#basic-usage

types/mirage-type-test.ts
```diff
@@ -0,0 +1,90 @@
...
+}
+
+server.db.loadData<IMovie>({
+  movies: [
+    { title: "Interstellar" },
+    { title: "Inception" },
+    { title: "Dunkirk" }
+  ]
+});
+server.db.loadData<IContact>({
```
zoltan-nz: Probably `loadData` can use generic type. What do you think?

types/index.d.ts
samselikoff: You're right, we need to add this

samselikoff:
This looks pretty awesome guys 👍Thanks so much for this work @zoltan-nz!

I'm unfamiliar with TS though it is high on my priority list. It looks like this makes typescript a dependency of Mirage. Does that add a lot to install size or affect non-TS users in other ways?

I'd also like to find other folks in the Ember ecosystem using Mirage with TS, and get their take on this. I know @chriskrycho has fiddled with this before.

chriskrycho:
Thanks for the ping! I'll carve out some time tomorrow to dig into these. And thanks @zoltan-nz for starting this work. cc. also @jamescdavis @dustinsoftware, both of whom are (I believe) working on apps making active use of Mirage with TS today.

types/index.d.ts
asantos00: Thanks! I've created a PR that adds this as esdoc types. WDYT?

https://github.com/miragejs/miragejs/pull/268

chriskrycho:
Hey folks, I need to punt this to slightly later in the week. I will say that my strongest recommendation is: we need *type tests* for this, using [tsd](https://github.com/SamVerschueren/tsd). We want to make sure that the API actually matches what we say it is, and we also want to make sure that the types shipped don't suddenly widen in ways that allow illegal things between versions.

(I'm *hopefully* going to be working on some standard infrastructure to make this easy over the course of the year, but this could be a great testbed for figuring out what makes for a good experience with that.)

chriskrycho:
For folks following along at home, when I *do* dig into this (🤞 hopefully Friday 🤞) I'm going to be referring to the following type definitions that @dfreeman, @jamescdavis, and I developed independently over the last few years, so that what we end up with is hopefully a best-of approach that includes the work done by @zoltan-nz, @asantos00, and the three of us:

- [@dfreeman's types](https://gist.github.com/dfreeman/33fc80164c0ad91d5e9480a94aa6454c)
- [ember-osf-web](https://github.com/CenterForOpenScience/ember-osf-web/tree/develop/types/ember-cli-mirage) (by @jamescdavis)
- [my types](https://gist.github.com/chriskrycho/f7d5d94302ca1fb5b99dc35e8094fd9d) (thanks to the folks at Olo for making them available for this purpose; I've sanitized them of everything in them that was specific to Olo)

dfreeman:
Awesome, thanks for pulling all this together @chriskrycho! I just updated that gist to include our type tests as well—they're written against `dtslint` rather than `tsd`, but I know you're familiar with reading that style too 😜

asantos00:
Maaan, this is huge. There are a lot of written types that I didn't knew about, which is amazing 🙏 . I think it's mostly a question of merging/choosing one. The hardest part will *definitely* be the types tests, which I'm not super aware of but will have a look

asantos00:
I was ready @dfreeman types and I think they're pretty complete. Do you think it makes sense to add them to this PR (or another one) so we can iterate on them?

jamescdavis:
Thanks for getting this going @zoltan-nz and @asantos00! This may not be relevant here because this is miragejs and no longer ember-specific, but something the [ember-osf-web types](https://github.com/CenterForOpenScience/ember-osf-web/tree/develop/types/ember-cli-mirage) do (that might be a bit crazy) is ensure that mirage factories and fixtures are type-compatible with the ember-data model they are based on. [Here's a fairly simple factory](https://github.com/CenterForOpenScience/ember-osf-web/blob/develop/mirage/factories/comment.ts) that makes use of this. You can see that it:
```ts
import Comment from 'ember-osf-web/models/comment';
```
defines an interface of traits:
```ts
export interface CommentTraits {
    withReplies: Trait;
    asAbuse: Trait;
}
```
and then passes their intersection as a type arg:
```ts
export default Factory.extend<Comment & CommentTraits>({
```
The type arg ensures that properties set in the Comment factory match type with the ember-data Comment model (or are a trait defined in CommentTraits).

To do this requires some [pretty interesting mapped types](https://github.com/CenterForOpenScience/ember-osf-web/blob/85e022d3663ba12a9dff5836872831d0b48e6a0b/types/ember-cli-mirage/index.d.ts#L37-L46). This may be overkill to try to support this for the first version of these types (and maybe not something we ever want to try to support in the "official" types), but just some food for thought.

zoltan-nz:
It looks to me, that `dtslint` behaves strangely. I tried it on macOS and inside a Linux docker container. Throw errors, like `TypeError: Cannot read property 'length' of undefined`

```
The 'member-access' rule threw an error in '/home/node/app/types/index.d.ts':
TypeError: Cannot read property 'length' of undefined
    at recur (/home/node/app/node_modules/dtslint/node_modules/tslint/lib/rules/memberAccessRule.js:91:57)
    at visitNodes (/home/node/app/node_modules/dtslint/node_modules/typescript/lib/typescript.js:18253:30)
    at Object.forEachChild (/home/node/app/node_modules/dtslint/node_modules/typescript/lib/typescript.js:18484:24)
    at recur (/home/node/app/node_modules/dtslint/node_modules/tslint/lib/rules/memberAccessRule.js:108:19)
    at visitNode (/home/node/app/node_modules/dtslint/node_modules/typescript/lib/typescript.js:18244:24)
    at Object.forEachChild (/home/node/app/node_modules/dtslint/node_modules/typescript/lib/typescript.js:18585:21)
    at recur (/home/node/app/node_modules/dtslint/node_modules/tslint/lib/rules/memberAccessRule.js:108:19)
    at visitNodes (/home/node/app/node_modules/dtslint/node_modules/typescript/lib/typescript.js:18253:30)
    at Object.forEachChild (/home/node/app/node_modules/dtslint/node_modules/typescript/lib/typescript.js:18486:24)
    at walk (/home/node/app/node_modules/dtslint/node_modules/tslint/lib/rules/memberAccessRule.js:89:15)
```

Also it complains about `var` keyword, however there isn't any `var` keyword...

```
Error: /home/node/app/types/index.d.ts:60:16
ERROR: 60:16    no-var-keyword                   Forbidden 'var' keyword, use 'let' or 'const' instead
```
Do you guys have any suggestion how could we test these types? It would be great to complete this PR, so `miragejs` can officially support Angular projects as well. ;)

samselikoff:
Kinda keeping one eye on this PR, I'm glad you all are collaborating and I don't want to be a bottleneck.

I still don't have TS experience but this is a big PR. I know some projects have a separate `@types` repo that contains the types.

Would it be helpful if we moved this to an official `@miragejs/types` repo? Would it help you all iterate + discuss it better?

jamescdavis:
@samselikoff keeping the types in the same repo is the best option. The `@types/` prefix is owned by Microsoft. Those packages are types that live in [DefinitelyTyped](http://definitelytyped.org/) and maintained completely separate from the packages the types are for. By default, TypeScript only knows how to access types in `@types/` packages or within the package that the types are for (defaulting to `index.d.ts` in the package root, but it can look elsewhere within the package via the `types` field in `package.json`). You *could* publish `@miragejs/types`, but TypeScript would not know to look there for miragejs's types.  It's much better to publish types within packages because they will always be compatible for the package version (assuming they are maintained). `@types/` types can only match major versions and thus can include types for features that don't exist in earlier minor versions. Also, being separate packages that are not controlled by package maintainers, coordinating types changes for new features and releases can be very difficult.

All this being said, the *best* best option is to write packages *in* TypeScript and generate type definitions automatically, but that is obviously a much larger ask of the package author (but something to consider for the future :wink:).

samselikoff:
Haha for sure, I absolutely think TS would be a great fit for the Mirage codebase. Right now I'm focused on the homepage for miragejs.com and making some educational material, to spread the word about miragejs. If and when there is more adoption it will justify more work on the project, including TS adoption.

Your points make sense! Thanks for the details reply 👍

dfreeman:
> Do you guys have any suggestion how could we test these types? It would be great to complete this PR, so miragejs can officially support Angular projects as well. ;)

@zoltan-nz I pulled your branch, and running dtslint with `--expectOnly` only gives one error (and only with `typescript@next`), which looks like a legitimate mismatch between the declarations and what one of the tests is trying to do.

I think `--expectOnly` makes sense for the kind of testing we're doing here, but it may also make sense to move those tests over to work against `tsd`, since that project seems to be a bit more general-purpose and actively maintained (while dtslint is mostly just intended as a tool for DefinitelyTyped)

zoltan-nz:
@dfreeman How about adding types and testing only with `dtslint` in this pull request, so types can be added now and @samselikoff  can release a new `miragejs` package with TypeScript support, meanwhile in a separated PR, we can start migrate to `tsd`?

If you guys have a sec, could you please check this PR and approve if you are happy with it? cc @chriskrycho @asantos00 I really appreciate your help and suggestions.

chriskrycho:
I second the decision to land with these tests and refactor next.

I'll review on Monday (since it's a holiday for me)!

types/index.d.ts
```diff
@@ -0,0 +1,369 @@
...
+    ModelDefinition,
+    ModelInstance
+  } from "miragejs/-types";
+  export { Server } from "miragejs/server";
+  export { Registry, ModelInstance } from "miragejs/-types";
+
+  /**
+   * A fake HTTP request
+   */
+  export class Request {
```
asantos00: Can't we reuse the `Request` from pretender? https://github.com/pretenderjs/pretender/blob/master/index.d.ts#L47

asantos00: There should be more reusable stuff too

types/index.d.ts
```diff
@@ -0,0 +1,369 @@
...
+
+    /** Handle a DELETE request to the given path. */
+    del(
+      path: string,
+      handler?: RouteHandler<Registry>,
+      options?: HandlerOptions
+    ): void;
+
+    /** Pass through one or more URLs to make real requests. */
+    passthrough(...urls: string[]): void;
```
asantos00: This was updated in last version, it can now get a string but can also get a `function` where the request object is injected

types/index.d.ts
```diff
@@ -0,0 +1,369 @@
...
+      Data extends Partial<Init>
+    >(modelName: K, count: number, data?: Data): Array<Init & Data>;
+
+    /** Handle a GET request to the given path. */
+    get(
+      path: string,
+      handler?: RouteHandler<Registry>,
+      options?: HandlerOptions
+    ): void;
+
```
asantos00: `Server` also has the `options` method (for the `OPTIONS` request in HTTP)

types/index.d.ts
```diff
@@ -0,0 +1,369 @@
...
+    /** Whether or not Mirage should log all requests/response cycles. */
+    logging: boolean;
+
+    /** A default number of ms to artifically delay responses for all routes. */
+    timing: number;
+
+    /** A default prefix applied to all subsequent route definitions. */
+    namespace: string;
+
```
asantos00: `urlPrefix` is missing here

types/index.d.ts
```diff
@@ -0,0 +1,369 @@
...
+      path: string,
+      handler?: RouteHandler<Registry>,
+      options?: HandlerOptions
+    ): void;
+
+    /** Pass through one or more URLs to make real requests. */
+    passthrough(...urls: string[]): void;
+
+    /** Load all available fixture data matching the given name(s). */
+    loadFixtures(...names: string[]): void;
```
asantos00: Even though this method exists, there's also the new `seeds` method that does the same with (imho) more explicit naming, better to keep both for now.

The method `routes` was also added in the meantime

types/index.d.ts
```diff
@@ -0,0 +1,369 @@
...
+     */
+    create<
+      K extends keyof Registry,
+      Init extends Registry[K],
+      Data extends Partial<ModelInitializer<Init>>
+    >(
+      modelName: K,
+      data?: Data
+    ): Init & { [K in keyof Init & keyof Data]: Exclude<Init[K], undefined> };
+
```
asantos00: `update` method missing

types/index.d.ts
zoltan-nz: I'm not sure how could you reuse, any idea?

types/index.d.ts
zoltan-nz: `seeds` and `routes` are not really listed in the documentation. https://miragejs.com/api/classes/server#load-fixtures

It is mentioned in the comment section of `loadFixtures`, but not really clear how could you use it.

As I see in the source code, `routes` and `seeds` can be a property on the `config` object.

https://github.com/miragejs/miragejs/blob/master/lib/server.js#L222-L236

types/index.d.ts
zoltan-nz: I don't see any `update` method in the `Schema` documentation .
https://miragejs.com/api/classes/schema

types/index.d.ts
zoltan-nz: Thanks. Added.

types/index.d.ts
zoltan-nz: Thanks, added.

types/index.d.ts
zoltan-nz: Thanks, added.

types/index.d.ts
zoltan-nz: Anyway, methods added without description.

types/index.d.ts
asantos00: Oh, you are right. I was misleading this with `schema.db.todos.update` for instance

types/index.d.ts
asantos00: Oh, I thought `pretender` was exporting the FakeXMLHttpRequest type, which it is not :( Never mind then

types/index.d.ts
zoltan-nz: No worries. Yep, it comes from the `model`, so it should work... ;)

types/index.d.ts
zoltan-nz: Maybe there is some way to reuse them, it would be great. Let's figure it out later. ;) Thanks for the suggestion. Definitely an important area where we can improve the type def further.

zoltan-nz:
@asantos00, thanks for reviewing.

@chriskrycho  any other suggestion?

@samselikoff what do you think, can we merge this to master and add to an official release? I would like to replace Pretender with Mirage.js in our app at our workplace. I also keen to create a sample Angular app what you can add as an example to the documentation. 

I really appreciate your help guys. Enjoy the weekend. :)

bboydflo:
> I also keen to create a sample Angular app what you can add as an example to the documentation.

that would be nice!

chriskrycho:
Would love it if at least *one* of @dfreeman, @jamescdavis, or myself could give it a 👍 before we merged. These types are *super* critical to the ecosystem, so worth one last check. I unfortunately spent all day today writing an Ember RFC instead (publishing on Monday), but will allocate time to this on Monday.

dfreeman:
I think my own fingerprints are on these types enough that a review from @chriskrycho or @jamescdavis would likely go further, but if y'all aren't able to take a look next week I'm happy to.

drmohundro:
Apologies for jumping into an ongoing conversation, but I'm hoping I'm either missing something or contributing a helpful comment. I pulled this down to try it out with some of my existing TypeScript code and ran into a few issues with it. Doesn't the `Server` type need a `constructor` method definition to support syntax like so (via the getting started docs)?

```js
new Server({
  routes() {
    // ...
  },
})
```

It may be that there is another means of configuring the Server that I didn't notice right away because I was going through the main getting started docs. If so, please disregard! Thanks for a great project!

smkamranqadri:
Hi guys,

I built sample application in angular to demonstrate miragejs, might helpful or may be it can be added in doc for examples.

https://github.com/smkamranqadri/angular-mirage-sample

Also would love to contribute so let me know if I can help in any way I can.

zoltan-nz:
@drmohundro good point. Thanks for your message.

So, I just added a `ServerConfig` interface to support constructor options.
https://github.com/miragejs/miragejs/blob/master/lib/server.js#L190

However, `this.get()` and similar methods are still not supported inside the `routes()` function. I'm not sure how should we solve this in TypeScript. Probably better at this stage, if you create the Server using `baseConfig()` as it is described in the API.
https://miragejs.com/api/classes/server

types/index.d.ts
```diff
@@ -0,0 +1,383 @@
...
+    /** Shutdown the server and stop intercepting network requests. */
+    shutdown(): void;
+  }
+}
+
+declare module "miragejs/db" {
+  /** The in-memory database containing all currently active data keyed by collection name. */
+  export default class Db {
+    [key: string]: unknown;
+  }
```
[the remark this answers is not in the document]
zoltan-nz: Thanks for your suggestion. 

Implemented.

types/index.d.ts
```diff
@@ -0,0 +1,495 @@
...
+    scenarios?: (server: Server) => void;
+
+    routes?: (this: Server) => void;
+    baseConfig?: (this: Server) => void;
+    testConfig?: (this: Server) => void;
+
+    inflector?: object;
+    identityManagers?: IdentityManager;
+    models?: any;
+    seralizers?: any;
```
NonPolynomial: typo
it is `serializers` not `seralizers`

types/index.d.ts
```diff
@@ -0,0 +1,495 @@
...
+  export type ModelInstance<Data = {}> = Data & {
+    id?: string;
+    attrs: Record<string, unknown>;
+    modelName: string;
+
+    /** Persists any updates on this model back to the Mirage database. */
+    save(): void;
+
+    /** Updates and immediately persists a single attr on this model. */
+    update<K extends keyof Data>(key: K, value: Data[K]): void;
```
NonPolynomial: mirage also allows to [update multiple attrs](https://miragejs.com/docs/main-concepts/models#updating-and-deleting-models)

```ts
/** Updates and immediately persists a single or multiple attr(s) on this model. */
update(changes: Partial<Data>): void;
```

types/index.d.ts
```diff
@@ -0,0 +1,495 @@
...
+    ): void;
+
+    delete(
+      path: string,
+      handler?: RouteHandler<Registry>,
+      options?: HandlerOptions
+    ): void;
+
+    /** Pass through one or more URLs to make real requests. */
+    passthrough(urls: ((request: Request) => any) | string | string[]): void;
```
NonPolynomial: params are optional

types/index.d.ts
zoltan-nz: @NonPolynomial thanks, fixed.

types/index.d.ts
zoltan-nz: @NonPolynomial Thanks. I just extended the code with your suggestion.

types/index.d.ts
zoltan-nz: @NonPolynomial Cheers. Updated. ;)

types/tests/overview-test.ts
```diff
@@ -0,0 +1,126 @@
...
+    movie: Model.extend({
+      castMembers: hasMany()
+    }),
+    castMember: Model.extend({
+      movie: belongsTo()
+    })
+  },
+  routes() {
+    this.get("/movies/:id/cast-members", (schema, request) => {
+      const movie = schema.movies.find(request.params.id);
```
zoltan-nz: @jamescdavis @chriskrycho @dfreeman @asantos00  Hey Team, it would be great if this code would pass the type check. Hope you can share some advice.

https://travis-ci.org/miragejs/miragejs/jobs/644912662#L543

`Property 'movies' does not exist on type 'Schema<Record...`

zoltan-nz: `schema.db.movies.find()` works and using schema.db is in the documentation... https://miragejs.com/docs/main-concepts/orm#why-an-orm

NonPolynomial:
@zoltan-nz thanks for your work! 👍

samselikoff:
@zoltan-nz thanks for keeping up with this, looks like you've been chugging away!

@jamescdavis @dfreeman @chriskrycho what do we need to ship this? Definitely don't want to block + this is still out of my wheelhouse. 

If you have time would appreciate any more input!

castrog:
Very excited for this to land as we're just getting started with Mirage.js in a new TS codebase.  We very much appreciate all the hard work that's gone into getting TypeScript support into the codebase!

zoltan-nz:
@samselikoff @jamescdavis @dfreeman @chriskrycho Hi Team, how about merging this PR to master and release it, so devs can start using it and we can improve it further in agile way, based on feedbacks and feature requests?

Additionally, I'm happy to start to migrate tests to `tsd`. What do you think?

samselikoff:
@zoltan-nz Happy to merge but I'd like one more once-over from a second pair of eyes - @jamescdavis @dfreeman or @chriskrycho anyone have time?

chriskrycho:
Will look at it first thing tomorrow morning. :) Sorry for the delay, folks!

samselikoff:
No problem at all + I really appreciate it!!

jamescdavis:
I should be able to take a look tomorrow as well!

types/index.d.ts
```diff
@@ -0,0 +1,496 @@
...
+    inflector?: object;
+    identityManagers?: IdentityManager;
+    models?: any;
+    serializers?: any;
+    factories?: any;
+
+    pretender?: PretenderServer;
+  }
+
+  export class Server<Registry = Record<string, ModelInstance>>
```
chriskrycho: Should this be constrained as to what it extends, or no? If so:

```suggestion
  export class Server<Registry extends Record<string, ModelInstance> = Record<string, ModelInstance>>
```

types/index.d.ts
```diff
@@ -0,0 +1,496 @@
...
+  }
+}
+
+declare module "miragejs/db" {
+  import DbCollection from "miragejs/db-collection";
+  import IdentityManager from "miragejs/identity-manager";
+
+  /** The in-memory database containing all currently active data keyed by collection name. */
+  export default class Db {
+    [key: string]: DbCollection | any;
```
chriskrycho: `| any` here makes the `DbCollection` part meaningless: you're saying the type is "something specific, or anything", which is the same as saying "anything". Likely it should just be `DbCollection`.

types/index.d.ts
```diff
@@ -0,0 +1,496 @@
...
+    createCollection(name: string, initialData?: any[]): void;
+    dump(): void;
+    emptyData(): void;
+    loadData(data: any): void;
+  }
+}
+
+declare module "miragejs/db-collection" {
+  import IdentityManager from "miragejs/identity-manager";
+  export default class DbCollection {
```
chriskrycho: We may want to circle back and make this generic so that it can wrap over a specific type—that can come in a follow-on pass, but note that doing so *will* technically constitute a breaking change to the types, so we may want to do it here and now.

```ts
export default class DbCollection<T> {
  constructor(
    name: string,
    initialData: T[],
    identityManager?: IdentityManager
  );

  all(): T[];

  // etc.
```

Note that this will likely also involve plumbing that info back through the rest of the system.

Given that would be a breaking change, we may want to land this particular PR, *then* do that work, but make sure to land it *before* cutting a Mirage release with the type defs published.

types/index.d.ts
```diff
@@ -0,0 +1,496 @@
...
+      modelName: K,
+      data?: Data
+    ): Init & { [K in keyof Init & keyof Data]: Exclude<Init[K], undefined> };
+
+    /** Locates one or more existing models of the given type by ID(s). */
+    find<K extends keyof Registry>(type: K, id: string): Registry[K] | null;
+    find<K extends keyof Registry>(
+      type: K,
+      ids: string[]
+    ): Collection<Registry[K]>;
```
chriskrycho: Can this ever return `null`? If so, this should be:

```suggestion
    ): Collection<Registry[K]> | null;
```

If it just always returns an empty collection, please ignore this comment! 👍

types/index.d.ts
```diff
@@ -0,0 +1,496 @@
...
+    ): Registry[K];
+
+    /** Locates an existing model of the given type by attribute value(s), if one exists. */
+    where<K extends keyof Registry>(
+      type: K,
+      attributes: Partial<Registry[K]>
+    ): Collection<Registry[K]>;
+    where<K extends keyof Registry>(
+      type: K,
+      // tslint:disable-next-line:unified-signatures
```
chriskrycho: Is this `tslint:disable-next-line` actually required?

types/index.d.ts
```diff
@@ -0,0 +1,496 @@
...
+    /** Returns the first model instance found of the given type */
+    first<K extends keyof Registry>(type: K): Registry[K] | null;
+  }
+}
+
+declare module "miragejs/test-support/setup-mirage" {
+  /** Enables Mirage in the current test scope. */
+  export default function setupMirage(hooks: {
+    beforeEach: () => void;
+    afterEach: () => void;
```
chriskrycho: QUnit's hooks values technically pass `assert` to these functions, so we should do likewise here. It's rare that those should be used, but possible. It likely makes sense to import the relevant bits from the QUnit and/or ember-qunit types, so these can just stay in sync with them.

types/tests/db-test.ts
```diff
@@ -0,0 +1,39 @@
...
+myDb.users.where({ name: "Link" });
+myDb.users.insert({});
+myDb.users.insert([]);
+myDb.users.firstOrCreate({ name: "Link" });
+myDb.users.remove();
+myDb.users.remove(1);
+myDb.users.remove({ name: "Zelda" });
+myDb.users.update({ name: "Ganon" });
+myDb.users.update(1, { name: "Young Link" });
+myDb.users.update({ name: "Link" }, { name: "Epona" });
```
chriskrycho: When you come back with the follow-on PR with test assertions, let's assert what we expect these types to be!

types/tests/model-test.ts
```diff
@@ -0,0 +1,35 @@
...
+people.models.map(model => {
+  model.id; // $ExpectType string | undefined
+  model.name; // $ExpectType string
+  model.modelName; // $ExpectType string
+  model.attrs; // $ExpectType Record<string, unknown>
+  model.foo; // $ExpectError
+
+  model.save();
+  model.reload();
+  model.destroy();
```
chriskrycho: We should assert that these are all `void` IIRC.

types/tests/relationships-test.ts
```diff
@@ -0,0 +1,88 @@
...
+>;
+declare const schema: Schema<PersonRegistry>;
+
+const people = schema.all("person");
+
+people.length; // $ExpectType number
+people.modelName; // $ExpectType string
+people.models.map(model => {
+  model.parent!.name; // $ExpectType string
+  model.parent!.parent!.name; // $ExpectType string
```
chriskrycho: You might want to invert this and make it `model.parent?.name` expecting type `string | null` (or `string | undefined`, whichever is correct) and the same for the two-parent lookup. We strongly prefer to *handle* optionals rather than to *assert over them*. Here it doesn't matter so much, as this still correctly tests the types, so you don't have to change it, but it'd be my recommendation. Code like this tends to get read by the curious and used as an example.

types/tests/server-test.ts
```diff
@@ -0,0 +1,63 @@
...
+  this.del("/foo");
+
+  this.passthrough("/_coverage/upload");
+  this.passthrough(request => request.queryParams.skipMirage);
+
+  this.loadFixtures();
+  this.seeds(this);
+  this.routes();
+
+  this.shutdown();
```
chriskrycho: As above, I suggest we add tests for the expect result type for each of these.

types/index.d.ts
```diff
@@ -0,0 +1,496 @@
...
+  export type BelongsTo<Registry> =
+      & (<K extends keyof Registry>(key?: K, options?: RelationshipOptions) => () => Registry[K] | undefined)
+      & (<K extends keyof Registry>(options: RelationshipOptions & { polymorphic: true }) => () => Registry[K] | undefined);
+
+  /** Declares a one-to-one relationship to another Mirage model type. */
+  export function belongsTo<T = ModelInstance>(
+    key?: string,
+    options?: RelationshipOptions
+  ): () => T | undefined;
+  export function belongsTo<T = ModelInstance>(
```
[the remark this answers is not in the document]
zoltan-nz: Thanks, suggestion applied.

types/index.d.ts
```diff
@@ -0,0 +1,496 @@
...
+  }
+
+  /** The registry-aware type of Mirage's `belongsTo` function */
+  // prettier-ignore
+  export type BelongsTo<Registry> =
+      & (<K extends keyof Registry>(key?: K, options?: RelationshipOptions) => () => Registry[K] | undefined)
+      & (<K extends keyof Registry>(options: RelationshipOptions & { polymorphic: true }) => () => Registry[K] | undefined);
+
+  /** Declares a one-to-one relationship to another Mirage model type. */
+  export function belongsTo<T = ModelInstance>(
```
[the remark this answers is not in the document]
zoltan-nz: Thanks, suggestion applied.

types/index.d.ts
```diff
@@ -0,0 +1,496 @@
...
+      ? FlattenRelationships<Data>
+      : {}
+    : {};
+
+  // Extracts factory definition info for the given key, if a corresponding factory is defined
+  type ExtractFactoryData<Factories, K> = K extends keyof Factories
+    ? Factories[K] extends FactoryDefinition<infer Data>
+      ? FlattenFactoryMethods<Data>
+      : {}
+    : {};
```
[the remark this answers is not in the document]
zoltan-nz: @dfreeman what do you think?

types/index.d.ts
zoltan-nz: You right, it is better if there is a type constrain. Thanks.

types/tests/db-test.ts
zoltan-nz: Thanks, good call. Assertions added.

types/tests/relationships-test.ts
zoltan-nz: TypeScript updated to 3.8, minimum required TS is also updated to 3.7 in dts header. Now we can use `?` syntax, so the test is updated. Thanks for the suggestion.

types/tests/server-test.ts
zoltan-nz: Thanks. Assertions added. ;)

types/tests/model-test.ts
zoltan-nz: Assertions added.

zoltan-nz:
@samselikoff it looks the new version of `dtslint` would expect Node.js `>=v12`. Are you planning to increase Node.js version on Travis maybe? If not I would downgrade `dtslint`.
https://travis-ci.org/miragejs/miragejs/jobs/653702607#L492

types/index.d.ts
zoltan-nz: Removing `any` would cause the following error, any suggestion how could we solve it?

```
Error: /Users/zoltan/projects/miragejs/types/index.d.ts:360:5
ERROR: 360:5  expect  TypeScript@next compile error:
Property 'createCollection' of type '(name: string, initialData?: any[] | undefined) => void' is not assignable to string index type 'DbCollection'.
ERROR: 361:5  expect  TypeScript@next compile error:
Property 'dump' of type '() => void' is not assignable to string index type 'DbCollection'.
ERROR: 362:5  expect  TypeScript@next compile error:
Property 'emptyData' of type '() => void' is not assignable to string index type 'DbCollection'.
ERROR: 363:5  expect  TypeScript@next compile error:
Property 'loadData' of type '(data: any) => void' is not assignable to string index type 'DbCollection'.
```

types/index.d.ts
zoltan-nz: According to the original library code, it can return null.
https://github.com/miragejs/miragejs/blob/master/lib/orm/schema.js#L239-L254
https://github.com/miragejs/miragejs/blob/master/lib/orm/schema.js#L597

types/index.d.ts
zoltan-nz: Good suggestion. Thanks. `@types/qunit` added and the above definition is extended with `assert: Assert`.

types/index.d.ts
zoltan-nz: The overload can be merged with the first declaration. Update committed.

types/index.d.ts
zoltan-nz: Good suggestion, probably it is worth to open an issue ticket to keep as a reminder.

samselikoff:
I think we could update node in Travis, it’s pretty unlikely folks will hit node errors with Mirage. I’m totally OK with that if it makes your life easier.

zoltan-nz:
Actually, I realised, the qunit method is not exist in the official API, so I removed it. Importing qunit type definition is also problematic, because different projects use different test runner (e.g jest).

types/index.d.ts
```diff
@@ -0,0 +1,496 @@
...
+  ): () => Collection<T>;
+  export function hasMany<T = ModelInstance>(
+    options?: RelationshipOptions
+  ): () => Collection<T>;
+}
+
+declare module "miragejs/-types" {
+  // Captures the result of a `Model.extend()` call
+  const ModelData: unique symbol;
+  interface ModelDefinition<Data = {}> {
```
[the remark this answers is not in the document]
dfreeman: Would `object` make more sense than `{}` as the constraint here? `{}` pretty much only rules out `null` and `undefined`:

http://www.typescriptlang.org/play/index.html#code/GYVwdgxgLglg9mABDAJgRgDwBVEFMAeUuYKAzogN4C+AfABT4BciWAlMzhQFCK+IBOuKCH5J8Abi5UuXVGjoBWVrPR1gAQwA2pXMrl0wITZr2rwKXMBhhcKU-IDkACxgP7dAMoBPALYAjOE06B2A4ODdWSX1qSJlQSFgEZBQAJmw8QmIyRDg-ACtcaHomFnYWSh4+QWFRRAkpGVQUxVNmjW1dFWbDY1a6c0trWz7nVz7vf0Dg0PDWWKa6GPEgA

types/index.d.ts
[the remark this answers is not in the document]
dfreeman: The conditional check has to happen _somewhere_, because there might be entities that only have a model but not a factory defined, or vice versa. The overall type info for a Mirage record is determined by [the combination of the two](https://github.com/miragejs/miragejs/pull/262/files#diff-88bec6beae84369c0376b56b3bb88fe1R176-R178).

types/index.d.ts
dfreeman: This is a weird method, in that it can take a single ID or an array of them. If it receives an array, it either returns a Collection containing all the requested items or raises an error, so you'll never get back `null` for this overload.

types/index.d.ts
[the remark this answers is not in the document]
zoltan-nz: @dfreeman should we change something here?

zoltan-nz:
Should we add or change anything in this PR before merging?

maciejmyslinski:
`this.resource` is not covered 🙈 
I'd recommend handling that in the next PR, though.
BTW I couldn't find `this.resource` mentioned in the API documentation neither.

WonderPanda:
I know that +1 style comments can be a pain some times but I just wanna echo that I'm eagerly waiting for this to get merged. MirageJS looks amazing and I want to start adopting it into multiple code bases but it's hard when it's the only part of my front end stack that isn't typed

samselikoff:
@WonderPanda Understood and no worries, it's helpful to know who all is waiting on this.

This PR is unique in that I am basically unsuited to be responsible for it, as TS is currently so far outside of my domain. I know it sucks to wait but Dan and Chris are the folks I've chosen to help audit Zoltan's fantastic work, and so we'll just have to wait until they can carve out some time.

I believe Chris' last comment was to get some input from Dan.

In the meantime, if you did want to start using this, you could give this PR a dry run in your project. You might need to run `yarn build` after installing it from github since there is a build step. Let me know if you need help.

dfreeman:
@chriskrycho and I connected on this the other day and should hopefully be able to get things moving again shortly!

glinesbdev:
I'm also waiting on this PR. Feel free to reach out if there is help needed here.

Marcus-Rise:
I'm too waiting types for this package

types/index.d.ts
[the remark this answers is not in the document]
dfreeman: Per my comment above, I don't think so, but @chriskrycho may have another suggestion.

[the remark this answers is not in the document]
chriskrycho: Nope, this is reasonable, I just wanted to make sure! 😅

types/index.d.ts
```diff
@@ -0,0 +1,487 @@
...
+  type ModelInitializer<Data> = {
+    [K in keyof Data]: Data[K] extends Collection<infer M>
+      ? Collection<M> | M[]
+      : Data[K];
+  };
+
+  /**
+   * An interface to the Mirage ORM that allows for querying and creating records.
+   */
+  export default class Schema<Registry = Record<string, ModelInstance>> {
```
chriskrycho:
```suggestion
  export default class Schema<Registry extends Record<string, ModelInstance> = Record<string, ModelInstance>> {
```

types/index.d.ts
```diff
@@ -0,0 +1,487 @@
...
+  export default class Db {
+    [key: string]: DbCollection | any;
+
+    constructor(initialData: [], identityManagers?: IdentityManager[]);
+
+    createCollection(name: string, initialData?: any[]): void;
+    dump(): void;
+    emptyData(): void;
+    loadData(data: any): void;
+  }
```
chriskrycho: This ends up pretty gnarly, but it does the right thing by merging two distinct types as the type exported for `Db`:

```suggestion
  type DbLookup = {
    [key: string]: DbCollection;
  };

  class DbClass {
    constructor(initialData: [], identityManagers?: IdentityManager[]);

    createCollection(name: string, initialData?: any[]): void;
    dump(): void;
    emptyData(): void;
    loadData(data: any): void;
  }

  /** The in-memory database containing all currently active data keyed by collection name. */
  export const Db: DbClass;
  export type Db = DbClass & DbLookup;
  export default Db;
```

types/index.d.ts
chriskrycho: See comment below for how to resolve this – it was a bit of a tricky one!

types/index.d.ts
chriskrycho: Opened an issue [here](https://github.com/miragejs/miragejs/issues/380)!

chriskrycho:
Thank you for your patience – had a bunch of things block me from getting back to this, but I have at last! There's one last pair of small changes to make, and after that let's ship this thing!

types/index.d.ts
zoltan-nz: Thanks @chriskrycho Great. Let's merge the actual PR and after that we can migrate to generic DbCollection. Cool stuff. :)

samselikoff:
🚀 Amazing team work guys!!! So happy to see you all come together on this.

Tried to get you three in as co-authors on the commit and it looks like it worked. Thanks again for your work on this – I know tons of people can't wait to try it!

I'll cut a release very soon 👍

zoltan-nz:
@samselikoff Oh, wow. You are not joking... ;) (It is already 1st of April in New Zealand.)

I'm sure there gonna be a few follow up update and refinement. I try to push them through now. Keep you posted.

samselikoff:
Haha no joking here! 

Anyway we don’t have to cut a release till you feel good, but I saw ship it so I shipped!

zoltan-nz:
Linter started to complain on the upcoming version of TypeScript. I just opened a ticket and let's solve it quickly. https://github.com/miragejs/miragejs/issues/382

jamescdavis:
These types don't typecheck, even on 3.8:
```
types/index.d.ts:214:20 - error TS2344: Type 'Registry' does not satisfy the constraint 'Record<string, { id?: string; attrs: Record<string, unknown>; modelName: string; save(): void; update<K extends never>(key: K, value: {}[K]): void; update(changes: Partial<{}>): void; destroy(): void; reload(): void; }>'.

214     schema: Schema<Registry>,
                       ~~~~~~~~
```

please delay shipping until #383 (which I'm reviewing now).

PR #268, opened 2020-01-07
docs: Add server options types

asantos00:
- Adds esdoc types for `Server` constructor arguments

Related https://github.com/miragejs/site/pull/199

If both are merged, docs will be as shown on the link ☝️

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -182,6 +182,27 @@ function extractRouteArguments(args) {
   @public
 */
 export default class Server {
+  /**
+   * Creates a Server
+   * @param {Object} options Server's configuration object
+   * @param {String} options.urlPrefix The base URL for the routes. Example: `http://miragejs.com`.
+   * @param {String} options.namespace The default namespace for the `Server`. Example: `/api/v1`.
+   * @param {Number} options.timing Default latency for the routes to respond to a request.
+   * @param {String} options.environment Defines the environment of the `Server`.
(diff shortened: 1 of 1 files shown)


chriskrycho:
This is a good addition to the other PR, *but* if the other PR is merged we want to remove all the ESDoc *type* annotations from this, as they're just noise at that point and accidentally easy to get out of sync.

asantos00:
Yes, completely agree with you on that. 

In order to generate the documentation on the website (as shown on the link) we'll have to change the documentation generator to generate from TypeScript instead of from `esdoc` and we'll also have to add type descriptions that it currently does not have.

Am I missing something here?

chriskrycho:
I know @dfreeman had done some work a year or more ago to get ESDoc or similar able to pull type info from TS and merging that with ESDoc or similar, for precisely this reason, but I don’t know what the state of that work is.

chriskrycho:
Following up on my above comment: after thinking on this through the week, landing the ESDoc types is valuable in its own right since we don't know how long it'll take to get to a point where we can pull this out from TS.

asantos00:
All right, and yes. I also believe it can be a good *temporary* solution until we find the ideal one that is the one you suggested.

Can you have a look to the types and give your thought on it?

Thanks @chriskrycho

samselikoff:
I think is a definite improvement over what we have today, and easy to change once TS types land!

PR #281, opened 2020-01-14
Fix passthrough function with queryParams

asantos00:
# Bug

Currently, passing a function to the `passthrough` method will break if the "request to be passthrough" has queryParams.

This PR fixes it

### DIFF
--- a/__tests__/external/browser-only/passthrough-test.js
+++ b/__tests__/external/browser-only/passthrough-test.js
@@ -157,12 +157,14 @@ describe("External | Browser only | Passthrough", () => {
     server.config({
       routes() {
         this.passthrough(request => {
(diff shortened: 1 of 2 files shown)


samselikoff:
Great find + fix, thank you!!

sakhisheikh:
Still can't access in passthrough function it says undefined. 
If I pass skipMirage I says skipMirage of undefined, 

`this.passthrough(request => {
  return request.queryParams.skipMirage;
});`

At the same I can access queryParams in route handlers.

samselikoff:
@sakhisheikh believe this is a different thing, I've opened up a new issue about it https://github.com/miragejs/miragejs/issues/359

Issue #396, opened 2020-04-07
Using mirage of versions above 0.1.35 fails to build and run


andSubmarine:
We are a team that is working on a Nuxt prototype and have been using mirage for this with TypeScript for about 2-4 weeks. Ever since version 0.1.36 was released then node have been unable to build and run our application. The only fix was reverting to an older version of mirage. One error can be seen below:

```
1:34 Module '"miragejs"' has no exported member 'JSONAPISerializer'.
  > 1 | import {Server, Model, Response, JSONAPISerializer} from 'miragejs'
      |                                  ^
```

My question is this: Have the way we should use mirage changed with the new typescript support?  Is there any documentation to describe how we should use mirage with typescript? 

mirage.ts:
```
import {Server, Model, Response, JSONAPISerializer} from 'miragejs'
import users from "./fixtures/users"
import contacts from "./fixtures/contacts"

export function makeServer(environment =  "development") {
    return new Server({
        environment,
        fixtures: {
            users,
            contacts
        },
        models: {
            user: Model,
            contact: Model,
            smsVerification: Model
        },
        serializers: {
            application: JSONAPISerializer
        },
...
```

In the above code, the JSONAPISerializer is not recognized as being available from miragejs and the schema do not seem to contain any of our fixtures (i.e. users for instance is currently an array of json objects). Below is (some of) the packages defined for our node setup: 

package.json:
```
{
...
    "dependencies": {
        "axios": "0.19.2",
        "nuxt": "2.11.0",
        "ts-loader": "6.2.1",
        "typescript": "3.8.3",
        "vue": "2.6.11",
        ...
     },
    "devDependencies": {
        "@babel/core": "7.9.0",
        "@nuxt/typescript-build": "0.6.2",
        "@nuxtjs/eslint-config": "2.0.2",
        "@nuxtjs/eslint-module": "1.1.0",
        "@types/jest": "25.1.4",
        "@vue/cli-plugin-unit-jest": "4.3.0",
        "@vue/cli-service": "4.3.0",
        "@vue/test-utils": "1.0.0-beta.32",
        "babel-core": "7.0.0-bridge.0",
        "babel-eslint": "10.1.0",
        "babel-jest": "25.2.6",
        "eslint": "6.8.0",
        ...
        "jest": "25.2.7",
        "miragejs": "0.1.37",
        "ts-jest": "25.3.1",
        "vue-jest": "3.0.5"
      }
}
```

samselikoff:
[...]

PR #406, opened 2020-04-09
Simplify typed usage of relationships

dfreeman:
## Intro

Chatting with @samselikoff yesterday about possible paths forward for #396, I had an 'aha moment' around dealing with the typing for relationships between models. What's currently in place requires some boilerplate to work with TypeScript, in order to introduce registry-aware versions of the `belongsTo` and `hasMany` functions. It's also a bit fragile in the face of highly-interconnected sets of models, as you have to be careful not to trip the typechecker's circular type declaration alarms.

## Changes

This PR addresses both of those problems by reworking the types of relationships not to carry the full type of the model they're related to, but rather only the name of that model in the registry. The actual type is resolved on-demand when the key is referenced, eliminating the risk of the typechecker rejecting your setup due to unbounded type recursion.

The main upshot of this is that TypeScript users can just use `belongsTo` and `hasMany` directly from Mirage, rather than having to do a whole song and dance of defining their registry type and doing `export const belongsTo = Mirage.belongsTo as Mirage.BelongsTo<Registry>;`. If a model name is specified for a relationship that isn't in the instantiating registry, the resulting type will be `unknown`.

## Implementation Details

While the details of this don't matter to the consumer, from the implementation perspective, a composed `Registry` type goes from looking like this:

```ts
type MyRegistry = {
  pet: ModelInstance<{
    name: string;
    owner: ModelInstance<{ name: string; pets: ... }>;
  }>;
  person: ModelInstance<{
    name: string;
    pets: Collection<ModelInstance<{ name: string; owner: ... }>;
  }>;
};
```

To this:

```ts
type MyRegistry = {
  pet: {
    name: string;
    owner: BelongsTo<'person'>;
  };
  person: {
    name: string;
    pets: HasMany<'pet'>;
  }
};
```

And previously, methods that would return `Registry[ModelName]` now return `Instantiate<Registry, ModelName>`, which builds out the appropriate `ModelInstance<...>` type, resolving relationships as necessary.

## Notes

This PR is broken up across two commits, and it'll likely be easier to grok the change looking at them separately. The first introduces an `AnyRegistry` type alias so that we don't have to keep typing `extends Registry<Record<string, ModelDefinition>, Record<string, FactoryDefinition>>` everywhere. The second actually makes the shift to on-demand resolution of relationship types.

/cc @chriskrycho @jamescdavis @zoltan-nz

### DIFF
--- a/types/tests/relationships-test.ts
+++ b/types/tests/relationships-test.ts
@@ -1,26 +1,19 @@
-import {
-  belongsTo,
-  BelongsTo,
-  hasMany,
-  HasMany,
-  Model,
-  Registry,
-} from "miragejs";
+import { belongsTo, hasMany, Model, Registry } from "miragejs";
 import Schema from "miragejs/orm/schema";
-
-const registryBelongsTo: BelongsTo<PersonRegistry> = belongsTo;
-const registryHasMany: HasMany<PersonRegistry> = hasMany;
 
 const PersonModel = Model.extend({
   name: "hello",
-  parent: registryBelongsTo("person"),
-  pets: registryHasMany("pet"),
-  friends: registryHasMany<"pet" | "person">({ polymorphic: true }),
+  parent: belongsTo("person"),
+  pets: hasMany("pet"),
+  friends: hasMany<"pet" | "person">({ polymorphic: true }),
+
+  nothing: belongsTo("nonexistent"),
+  muchAdoAboutNothing: hasMany("nonexistent"),
 });
 
 const PetModel = Model.extend({
   name: "fido",
-  owner: registryBelongsTo("person"),
+  owner: belongsTo("person"),
 });
 
 type PersonRegistry = Registry<
@@ -34,6 +27,9 @@ const people = schema.all("person");
 people.length; // $ExpectType number
 people.modelName; // $ExpectType string
 people.models.map((model) => {
+  model.nothing; // $ExpectType unknown
+  model.muchAdoAboutNothing; // $ExpectType Collection<unknown>
+
   model.parent?.name; // $ExpectType string | undefined
   model.parent?.parent?.name; // $ExpectType string | undefined
   model.pets.models[0].name; // $ExpectType string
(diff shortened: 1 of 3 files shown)


zoltan-nz:
LGTM 👍

zoltan-nz:
@samselikoff I think it is ready to ship, so the other issue could be closed as well.

chriskrycho:
Will gander at this either later today (iffy) or tomorrow (not so iffy)!

types/tests/relationships-test.ts
```diff
@@ -34,6 +27,9 @@ const people = schema.all("person");
 people.length; // $ExpectType number
 people.modelName; // $ExpectType string
 people.models.map((model) => {
+  model.nothing; // $ExpectType unknown
```
chriskrycho, line 30: (Thinking out loud.) This makes me wonder about the tradeoff—the types mean we end up with `unknown` here; we could also restructure so it resolves as `never`. `unknown` is certainly more flexible, and for a codebase in transition to TS likely more *accurate*, as some types won't be provided yet. Definitely a thing for a later point, but it might be interesting to supply an alternative set of imports and give users the ability to use imports to specify which variant they want – something like `import 'miragejs/strict';` with the types resolving to `never`. 🤔 Definitely, definitely does not need to be a change here, though.

samselikoff:
Awesome, thank you so much everyone!

types/tests/relationships-test.ts
dfreeman: If it resolves as `never`, isn't that effectively the same as resolving to `any` since it's in a covariant position? Unless we did a mapping pass that eliminated keys whose values were `never`, though the failure mode for that seems more confusing to me from the user perspective ("why does it say this key doesn't exist on my model? I defined it right there!")

We could introduce some kind of `InvalidModelName` sigil or something to resolve it to instead of `unknown`, if that seems like it would be more helpful in nudging people to understand what they did wrong 🤔

chriskrycho: Yeah, you’re right. And anyway you’d want the error to happen at the definition site, not the use site.

PR #431, opened 2020-04-23
Passthrough data urls requests

cdes:
This passes through data urls without raising exceptions.

For reference, see this pull request:
https://github.com/miragejs/site/pull/528

### DIFF
--- a/__tests__/external/browser-only/passthrough-test.js
+++ b/__tests__/external/browser-only/passthrough-test.js
@@ -186,6 +186,12 @@ describe("External | Browser only | Passthrough", () => {
       `Mirage: Your app tried to GET '/movies'`
     );
   });
+
+  test("it passes through data urls", async () => {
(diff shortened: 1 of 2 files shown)


ryanto:
Thanks!

Mirage currently ships with a couple [default passthroughs](https://github.com/miragejs/miragejs/blob/master/lib/server.js#L117), so we should probably add to that list rather than unhandled request.

I'd like the understand which passthroughs safely solve this issue. Does adding `this.passthrough('data:**')` in your server definition solve this? Or do we need to do regex on every request?

cdes:
@ryanto: passthrough behavior does not work for some reason.  That's why I needed to do what I did.

You can see it here:
https://codesandbox.io/s/mirage-data-urls-b3r3b?file=/src/index.js

kieranbop:
I've stumbled accross this, where the passthrough still throws excepts to specific endpoints.

PR #467, opened 2020-05-13
Introduce `createServer`

dfreeman:
@samselikoff we had a conversation about this a little over a month ago, and I went and implemented it but then never actually committed the files and opened a pull request 🤦 

This PR introduces a `createServer` function as an alternative to writing `new Server`. While it doesn't add any new functionality, it does have a couple of advantages:
 - It avoids having consumers of the library call a side-effecty constructor (whether that's a code smell or not is a matter of personal opinion, but it's certainly been the topic of lots of angsty blog posts)
 - It allows for type-safe interactions with the server in the context of your config, which was the last lingering topic of discussion in #396.
  In other words, if I write `schema.all('foo')` in a route handler, TypeScript can tell based on the `foo` model/factory I provide what type the result will have. Or, if I haven't provided a `foo` model, it can correctly flag `schema.all('foo')` as an error.

/cc @zoltan-nz @chriskrycho

### DIFF
--- a/__tests__/internal/unit/server-test.js
+++ b/__tests__/internal/unit/server-test.js
@@ -1,4 +1,5 @@
 import {
+  createServer,
   Server,
   Model,
   Factory,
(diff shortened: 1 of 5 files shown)


samselikoff:
Awesome!! Thank you so much for this, I know it will make TS users happy 🙏

@chriskrycho @zoltan-nz any chance y'all could take a quick look or give it a shot in your TS projects?

zoltan-nz:
@samselikoff Thanks for the heads up.

@dfreeman Brilliant improvement. Love it.

I give it a try today in my project, however it looks good to me already.

zoltan-nz:
@samselikoff I played with it and works like a charm. Thanks. :)

samselikoff:
Amazing – thanks everyone!

PR #515, opened 2020-06-16
Would be nice to have a findBy in a collection

m-soli:
specifically for hasMany associations - for instance in an ember data model
Model.hasMany('otherModel')

in actual code: Model.otherModels.findBy(something)

in testing mirage collection for hasMany throws an error that findBy is not a function ¯\_(ツ)_/¯

### DIFF
--- a/lib/orm/collection.js
+++ b/lib/orm/collection.js
@@ -261,7 +261,16 @@ export default class Collection {
 
     return new Collection(this.modelName, filteredModels);
   }
-
+  
+  /**
+    Returns a collection model by specified key.
+    @method findBy
+    @param {STRING} key
+    @param {any} value
+  */
+  findBy(key, value) {
+    return this.models.filter((m)=> m[key] === value)[0];
+  }
   /**
      Returns a new Collection with its models sorted according to the provided [compare function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#Parameters).
 


samselikoff:
Happy to add this! Think you need to run prettier on your code, and also we'll need to get a test in here before merging.

What do you mean by

> in testing mirage collection for hasMany throws an error that findBy is not a function

? Could you clarify a bit?

m-soli:
> Happy to add this! Think you need to run prettier on your code, and also we'll need to get a test in here before merging.
> 
> What do you mean by
> 
> > in testing mirage collection for hasMany throws an error that findBy is not a function
> 
> ? Could you clarify a bit?

I believe this change might be better suited for ember-mirage-cli.

I have a test case - where the model being tested (model a) has an association of has_many to another model (model b). 

There is a computed property that tries to find a record of model b like so:
```js
// (s) for pluralization
model_a.model_b(s).findBy('id', '2'); 
```

But since in testing the model is a mirage factory model - the hasMany returns a Collection object - which doesn't have a findBy method - so test fails on findBy is not a function of Collection.

There might be something I am missing - I am new with mirage.
Since this is ember specific I think it belongs in ember-mirage-cli repo. 
I am going to close this pull request. 

Please advice is this an issue? and if so should I create an issue for ember-mirage-cli ?
Thank you

samselikoff:
@m-soli Ah yes – give this section of the docs a read: https://www.ember-cli-mirage.com/docs/testing/integration-and-unit-tests#creating-ember-data-models-on-the-client-with-mirage

PR #525, opened 2020-06-23
Initial support of traits in Typescript

### LINKED ISSUE #524: Trait and association not available in Typescript
It is not currently possible to use `trait` or `association` in factories when using Typescript.

`association` missing can be worked around. But it would be really nice with access to `trait`.

I might give the typings a shot this week. But since I'm pretty new to this project, it might be significantly faster for someone else 😄

bjarkehs:
In attempt to fix #524 this adds a very simple form of support for traits in Typescript.

Currently missing any handling of trait properties.

Downside of this is that it currently does not handle the traits' added properties.

I attempted to type the trait function and `FlattenFactoryMethods` so I could filter out the trait and instead add the trait's properties. But I couldn't figure it out.

I attempted something like:

```typescript
// Extract factory method return values from a factory definition
type FlattenFactoryMethods<T> = {
  [K in keyof T]: T[K] extends (n: number) => infer V
    ? V
    : T[K] extends (data: {}) => Trait<infer U>
    ? U
    : T[K];
};

type Trait<T extends {} = {}> = T & { __isTrait__: true };
```

This is not part of the PR since I couldn't get it working.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -115,6 +115,9 @@ declare module "miragejs" {
   export function hasMany<K extends string>(
     options?: RelationshipOptions
   ): HasMany<K>;
+
+  /** Declares traits for Factory */
+  export function trait<Data>(data: Data): Data;
 }
 
 declare module "miragejs/-types" {
@@ -340,7 +343,11 @@ declare module "miragejs/server" {
       K extends keyof Registry,
       Init extends Instantiate<Registry, K>,
       Data extends Partial<Init>
-    >(modelName: K, count: number, data?: Data): Array<Init & Data>;
+    >(
+      modelName: K,
+      count: number,
+      ...data: (Data | keyof Data)[]
+    ): Array<Init & Data>;
 
     /** Handle a GET request to the given path. */
     get(
@@ -512,7 +519,7 @@ declare module "miragejs/orm/schema" {
       Data extends Partial<ModelInitializer<Init>>
     >(
       modelName: K,
-      data?: Data
+      ...data: (Data | keyof Data)[]
     ): Init &
       { [K in keyof Init & keyof Data]: Exclude<Init[K], undefined | null> };
 


samselikoff:
@dfreeman @chriskrycho @zoltan-nz @jamescdavis If any of you have any spare cycles to take a look at this I'd appreciate it!

If not feel free to ignore :)

samselikoff:
@bjarkehs thanks for opening this!

dfreeman:
I should have some time to look at this later this week—first I'll have to read up on how traits actually work 😅

samselikoff:
Oh man traits are the bomb. Similar to FactoryBot traits if you've ever used em.

https://miragejs.com/docs/main-concepts/factories/#traits

jamescdavis:
I'm hoping to look at this week as well. And I'm pretty familiar with traits. :wink:

dfreeman:
Just a quick update on this: I spent some time over the weekend puzzling out how to type the way traits work. It seems like it's possible, but... complicated 😄 

I'll coordinate with @jamescdavis and see if we can't get a cleaned up version put together to propose here

samselikoff:
Cool – feel free to propose new API if it helps too (like we did with createServer)

thedaviddias:
@jamescdavis @dfreeman any update in adding trait and association?

yekver:
really waiting for this PR!

MrLoh:
so am I any progress on maybe at least not throwing a type error for this anymore

rubenmoya:
Anything we can do to push this forward? Even without having types in the traits properties at least it won't throw an error when trying to use traits.

dfreeman:
I picked this up to work on it over the holidays and wound up a little stumped by the change that landed in https://github.com/miragejs/miragejs/pull/695 and what problem it was addressing. Had a quick conversation in the #typescript channel about it and Sam suggested checking with the author of the PR.

@NLincoln can you offer any insight into that fix? I see you linked to some example code that doesn't work because factories don't understand ES getters, but I don't quite understand how the types change addresses that. I'm also happy to chat on Discord if that would be easier than going back and forth on issue comments 🙂

josemarluedke:
Looking forward to having TypeScript support for traits!

types/index.d.ts
```diff
@@ -115,6 +115,9 @@ declare module "miragejs" {
   export function hasMany<K extends string>(
     options?: RelationshipOptions
   ): HasMany<K>;
+
+  /** Declares traits for Factory */
+  export function trait<Data>(data: Data): Data;
```
josemarluedke, line 120: Is there any way to also add type for `afterCreate` method?

wagenet:
@dfreeman it definitely seems incorrect to me as well. See https://github.com/miragejs/miragejs/pull/695#issuecomment-984193138

jasikpark:
Wouldn't 

```ts
/** Declares traits for Factory */
  export function trait<Data>(data: Data): Data & { __isTrait__: true };
```

be slightly more correct, since `trait` is just adding that value to `Data`?

PR #537, opened 2020-07-01
Update serializer documentation

alexdunne:
Use the imported serializer in the demo codeblock


samselikoff:
Wow pretty egregious! Haha sometimes I wonder how these slip through 🙈

Thanks for the fix!

PR #695, opened 2020-11-02
[TS] Basic support for functions in factories

NLincoln:
There's a small limitation of the current types in that we can't use functions within factory definitions. This fixes it so that we can.

This should fix https://github.com/miragejs/examples/issues/9 by making it so that we don't need to use the property-getter workaround

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -137,8 +137,14 @@ declare module "miragejs/-types" {
 
   // Captures the result of a `Factory.extend()` call
   interface FactoryDefinition<Data extends {} = {}> {
-    extend<NewData>(data: NewData): FactoryDefinition<Assign<Data, NewData>>;
+    extend<NewData>(
+      data: WithFactoryMethods<NewData>
+    ): FactoryDefinition<Assign<Data, FlattenFactoryMethods<NewData>>>;
   }
+
+  type WithFactoryMethods<T> = {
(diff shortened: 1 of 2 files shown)


samselikoff:
Thank you!

timbakkum:
Hi, when will this be released? Latest version is from september. 
We are migrating to TypeScript and use mirage factories for our unit tests as well, so we'd like to make them type-safe.

samselikoff:
@timbakkum Looks like master is breaking due to some failures with the TS type tests. If you have time and could look into those that would definitely help me get a release out faster!

wagenet:
I don't think this is correct. We have code like this:

```ts
export default Factory.extend({
  // This has to come before `id` so we have a value for `this.sha`
  sha(n: number): string {
    return `ABC${n}`;
  },

  id() {
    return `OirVnUtLCe_0:${this.sha}`;
  }
}
```

And previously, `this.sha` would report that it was a `string`. Now it claims to be `string | ((n: number) => string)` but when I add a debugger, it's still a `string`. It looks like the `this` is a special object that is created by iterating over each object in the hash passed to `Factory.extend` in order. This also means that the order of the properties matters as defining `id` will lead to the value of `this.sha` being `undefined` when `id` is called.

wagenet:
Ok, I see what's happening here now. I think this code is actually correct, by the `this` in the factory methods is not what is inferred.

PR #792, opened 2021-02-17
Fix - `db.createCollection()` returns an instance, not `void`

MrChocolatine:
Fix the type returned by the method `db.createCollection()`.

Currently it mentions that `void` is returned, however the actual instance of the db is returned:
https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/lib/db.js#L142

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -422,7 +422,7 @@ declare module "miragejs/db" {
   class DbClass {
     constructor(initialData: [], identityManagers?: IdentityManager[]);
 
-    createCollection(name: string, initialData?: any[]): void;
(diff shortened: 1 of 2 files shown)


MrChocolatine:
Dang, I am a bit stuck, I don't really know what to do with the remaining error:

```
Error: /home/travis/build/miragejs/miragejs/types/tests/server-test.ts:67:5
ERROR: 67:5  expect  TypeScript@4.0 expected type to be:
  Collection<ModelInstance<{}>>
got:
  Collection<Instantiate<Registry<AnyModels, AnyFactories>, "asfd">>
```

Any help please :)?

Issue #832, opened 2021-03-20
Limited Typing Exposure of Collection


nbibler:
The [Collection class](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/lib/orm/collection.js) seems to have several useful public methods exposed (like `includes`, `filter`, `sort`, `slice`, …) that aren't currently identified in the TypeScript typings:

https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/types/index.d.ts#L87-L94

Is it intentional that these Array-like utility methods are hidden from the typings? Are they not intended to be utilized outside of the package?

samselikoff:
Nope not intentional! I don't maintain the types so it looks like they're just incomplete. Feel free to add them!

nbibler:
Sounds good, I'll get a PR up later tonight. Just wanted to check before putting in the effort. 😁

nbibler:
I've opened PR #833 to add the missing method signatures. Closing this issue and delegating to the PR.

PR #833, opened 2021-03-21
Add Collection Method Typings

nbibler:
This PR adds TypeScript method typings to the Collection class.

See #832.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -88,9 +88,21 @@ declare module "miragejs" {
    * A collection of zero or more Mirage model instances.
    */
   export class Collection<T> {
+    add(model: T): Collection<T>;
+    destroy(): Collection<T>;
+    filter(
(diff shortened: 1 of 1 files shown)


samselikoff:
Looks good! I think the build is currently broken on master which I suspect is why this PR is failing. Someone will need to fix it before this PR can get merged.

I don’t currently have the time so if someone could look into it that’d be a huge help!

nbibler:
@samselikoff: I can attempt it. :)

It looks like the functional tests are all passing. The `lint:ts` check, however, seems to be waffling different failures between TypeScript 3.9 and TypeScript 4.0. I'll try to figure out what's going on there.

samselikoff:
Thank you!

IanVS:
Thanks for the PR.  It looks like https://github.com/miragejs/miragejs/pull/900 adds the same types, but includes tests and comments, so I'll close this PR and move forward with the other.  Thanks again, though!

PR #850, opened 2021-04-06
Support selected embedded relationships

### LINKED ISSUE #841: Embedding Some Relationships, but Not All
Is it possible to embed _some_ of the relationships, but not _all_ of them?

It seems like the `embed` serializer is an all-or-nothing type of deal, without allowing for some relationships to be embedded and others to be "normal". Weird as that requirement might be, it's a situation we're running into right now!

NoneOfMaster:
Closes #841

### DIFF
--- a/lib/serializer.js
+++ b/lib/serializer.js
@@ -299,7 +299,7 @@ class Serializer {
       }
       ```
 
-      Setting `embed` to true will embed related records:
+      Setting `embed` to true will embed all related records:
 
       ```js
       Serializer.extend({
@@ -323,8 +323,11 @@ class Serializer {
         }
       }
       ```
+
+      You can also define `embed` as a function so it can be determined dynamically.
     */
     this.embed = this.embed || undefined; // this is just here so I can add the doc comment. Better way?
+    this.embedFn = isFunction(this.embed) ? this.embed : () => !!this.embed;
 
     /**
       Use this to define how your serializer handles serializing relationship keys. It can take one of three values:
@@ -473,11 +476,8 @@ class Serializer {
     didSerialize = {},
     lookupSerializer = false
   ) {
-    let hash, serializer;
-
-    if (!lookupSerializer) {
-      serializer = this; // this is used for embedded responses
-    }
+    let serializer = this;
+    let hash;
 
     // PolymorphicCollection lacks a modelName, but is dealt with in the map
     // by looking up the serializer on a per-model basis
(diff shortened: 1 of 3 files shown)


NoneOfMaster:
CI build error looks unrelated:
```
Error: /home/travis/build/miragejs/miragejs/types/tests/server-test.ts:67:5

ERROR: 67:5  expect  TypeScript@4.0 expected type to be:

  Collection<ModelInstance<{}>>

got:

  Collection<Instantiate<Registry<AnyModels, AnyFactories>, "asfd">>
```

navels:
bump?

IanVS:
I rebased this PR, and I think the typescript failure is unrelated.  

I think the only thing that's missing from the PR is an improvement to our TS type here: https://github.com/miragejs/miragejs/blob/f68581a1b56c4d3e5efcee94641fa761fd3add60/types/index.d.ts#L639.

IanVS (approved):
CI is green, thanks for the contribution.

lib/serializer.js
```diff
@@ -323,8 +323,11 @@ class Serializer {
         }
       }
       ```
+
+      You can also define `embed` as a function so it can be determined dynamically.
     */
     this.embed = this.embed || undefined; // this is just here so I can add the doc comment. Better way?
+    this.embedFn = isFunction(this.embed) ? this.embed : () => !!this.embed;
```
navels, line 330: This doesn't actually work, right?  An ES6 derived serializer will not be able to set `this.embed` before this code runs, effectively making `this.embedFn = () => true;`

IanVS: I'm not sure I quite follow, @navels.  What do you mean by "An ES6 derived serializer"?

navels: Aha, this is only a problem when defining the `embed` function as a class property, which is an ES2022 thing.  E.g.,

```
class Serializer {
  constructor() {
    console.log(`typeof this.embed: ${typeof this.embed}`);
  }
}

class FooSerializer extends Serializer {
  /* classical class method declaration */
  embed(key) {
    return false;
  }
}

class BarSerializer extends Serializer {
  /* ES2022 class property form */
  embed = (key) => false;
}

new FooSerializer();
/* console output: "typeof this.embed: function" */

new BarSerializer();
/* console output: "typeof this.embed: undefined" */
```
the difference being in the latter case `this.embed` doesn't exist until after the constructor has run.

So we can also work around this by not using the ES2022 syntax, but still. Kinda interesting potential class of bugs that new syntax is going to expose.

PR #900, opened 2021-05-27
Update collection TS definitions

adc-mhaugen:
Add missing method TS definitions for Collection object.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -222,12 +222,10 @@ declare module "miragejs/-types" {
 
   // Returns the instantiated type of the given model if it exists in the
   // given registry, or `unknown` otherwise.
-  type InstantiateIfDefined<
-    Registry,
-    ModelName
-  > = ModelName extends keyof Registry
-    ? Instantiate<Registry, ModelName>
-    : unknown;
+  type InstantiateIfDefined<Registry, ModelName> =
+    ModelName extends keyof Registry
+      ? Instantiate<Registry, ModelName>
+      : unknown;
 
   // The type-level equivalent of `Object.assign`
   type Assign<T, U> = U & Omit<T, keyof U>;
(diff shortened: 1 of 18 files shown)


types/index.d.ts
```diff
@@ -88,9 +88,47 @@ declare module "miragejs" {
...
+    save(): Collection<T>;
+
+    /** Returns a new Collection with a subset of its models selected from begin to end. */
+    slice(begin: number, end: number): Collection<T>;
+
+    /** Returns a new Collection with its models sorted according to the provided compare function. */
+    sort(f: (a: T, b: T) => number): Collection<T>;
+
+    /** Updates each model in the collection, and immediately persists all changes to the db. */
+    update: <K extends keyof T>(key: K, val: T[K]) => Collection<T>;
```
dfreeman, line 131: [nit] can you use method syntax for this? There are some subtleties with how subtyping behaves based on how function/method fields are declared.

```suggestion
    update<K extends keyof T>(key: K, val: T[K]): Collection<T>;
```

dfreeman (approved):
Thank you, @adc-mhaugen!

types/tests/server-test.ts
```diff
@@ -64,7 +64,7 @@ new Server({
   routes() {
     this.namespace = "api";
 
-    this.schema.all("asfd"); // $ExpectType Collection<ModelInstance<{}>>
+    this.schema.all("asfd"); // $ExpectType Collection<Instantiate<Registry<AnyModels, AnyFactories>, "asfd">>
```
adc-mhaugen, line 67: @dfreeman Looks like the expected type here changed.  Is this correct?

adc-mhaugen: Hmmm, the test are now failing with this change, so perhaps I was too hasty.

IanVS:
@adc-mhaugen we've updated our CI infrastructure, would you be willing to rebase this PR on master?  I'd like to get it green and merged in, if possible.

IanVS:
Merging from upstream was all that was needed, especially since I squash-merged.  If we needed to keep the commits, and wanted a cleaner history, an actual rebase would be done by running `git refresh && git rebase upstream/master`.  But all I wanted was for CI to run, so merging master worked fine too.  :+1:

PR #947, opened 2021-07-03
fix:reassign the constructor of  MirageError

peterczg:


IanVS:
Hi, can you please add some details about why this change is necessary?

PR #977, opened 2021-08-30
Add type for head request handler

IanVS:
I found that miragejs handles `this.head` route handlers just fine, but the signature was missing from the types.  So this adds it.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -396,6 +396,13 @@ declare module "miragejs/server" {
       options?: HandlerOptions
     ): void;
 
+    /** Handle a HEAD request to the given path. */
+    head(
+      path: string,
+      handler?: RouteHandler<Registry>,
+      options?: HandlerOptions
+    ): void;
+
     /** Pass through one or more URLs to make real requests. */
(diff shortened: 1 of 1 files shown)


IanVS:
@samselikoff this is a pretty small and straightforward PR to update some typescript.  Mind taking a quick look when you have a chance?  Thanks!

samselikoff:
@IanVS happy to merge but looks like build is failing :(

I think it's been failing for a while & we haven't cut a release in a while. Any help fixing master would be the fastest way to get this in there!

IanVS:
Thanks @samselikoff, I did in fact push up a PR to fix the build failure, as well.  :). https://github.com/miragejs/miragejs/pull/984

samselikoff:
Invited you to the repo, so I don't block you in the future :) (No pressure to help out anywhere else of course). 

Thanks again! Should be able to merge after new tests run

samselikoff:
Ok why don't you pull the trigger!

samselikoff:
Can also run through how we release too, if you have the cycles. Would be really great to get this written down so I'm not a blocker anymore

IanVS:
👍 I sent you an email.

PR #984, opened 2021-09-23
Fix CI: Remove type assertion that fails in TS 4.0

IanVS:
It seems the types changed a little in TS 4.0, so I've removed the $ExpectType that has been causing CI to fail.


samselikoff:
Thank you Ian!

PR #988, opened 2021-10-01
Upgrade and lock-down prettier version

IanVS:
This is an upgrade of prettier from 2.2.1 to 2.4.1.  I also locked down the package.json, as prettier recommends, because any change to the version can cause formatting differences so it's best to be intentional about upgrades.

This PR also includes the results of running the new version of prettier, which made a few formatting changes, but not a ton.


samselikoff (approved):

IanVS:
I'm not sure why CI isn't running on this PR...

samselikoff:
Really strange, never seen this before. GH actions would be awesome! Travis is def just there for historical reasons.

image

Thanks so much for your help, if you know how to fix this on Travis (seems like its a bug or theres just a switch I have to flip somewhere to get more oss credits) I can help, but if you're down to get GH actions working go for it!!

IanVS:
I've opened https://github.com/miragejs/miragejs/pull/990 to add github actions for lint + test.

PR #989, opened 2021-10-01
Update dtslint to 4.1.6

IanVS:
This updates the version of dtslint from 4.0.6 to 4.1.6.

~~It also removes the direct dependency on typescript~~ Edit: That didn't work out, typescript does seem to still need to be installed for dtslint to run.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,8 +89,7 @@
     "rollup-plugin-multi-entry": "^2.1.0",
     "rollup-plugin-node-resolve": "^5.2.0",
     "tslint": "^6.1.1",
-    "tslint-config-prettier": "^1.18.0",
-    "typescript": "^4.4.3"
+    "tslint-config-prettier": "^1.18.0"
   },
   "engines": {
     "node": "6.* || 8.* || >= 10.*"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -90,9 +90,12 @@
...
     "tslint": "^6.1.1",
     "tslint-config-prettier": "^1.18.0",
-    "typescript": "^4.0.3"
+    "typescript": "^4.4.3"
   },
   "engines": {
     "node": "6.* || 8.* || >= 10.*"
+  },
+  "resolutions": {
+    "@definitelytyped/typescript-versions": "0.0.89"
```
IanVS, line 99: This was necessary to avoid the yarn error:

```
Pattern ["@definitelytyped/typescript-versions@latest"] is trying to unpack in the same destination "/Users/ianvs/Library/Caches/Yarn/v6/npm-@definitelytyped-typescript-versions-0.0.89-815ba4ec6e2e0486e0d192ba19d7359cd9610f38-integrity/node_modules/@definitelytyped/typescript-versions" as pattern ["@definitelytyped/typescript-versions@^0.0.89","@definitelytyped/typescript-versions@^0.0.89","@definitelytyped/typescript-versions@^0.0.89"]. This could result in non-deterministic behavior, skipping.
```

samselikoff (approved):

PR #990, opened 2021-10-01
Move to github actions for CI

IanVS:
Travis has run out of credits, and github actions are more than powerful enough to run linting and tests, and in fact we can run against all the versions of node that are supported, to ensure no breaking changes are made accidentally.  

I've left the release job in travis for now, we can figure out what to do with that before the next time we need to release.


IanVS:
Interesting, windows seems to be failing some of the tests.  I can remove windows for now, but it might be worth looking into at some point.

samselikoff (approved):

PR #996, opened 2021-10-01
Update transitive dependencies

IanVS:
Now that all of our direct dependencies are updated, this PR updates all of the transitive dependencies (sub-dependencies).  I did this by deleting the `yarn.lock` file, and re-creating it.  This allows semver to do its thing and de-duplicate as many dependencies as possible, which actually reduces the total number of packages installed in the mirage repo from 1024 to 960.

This isn't something I'd recommend doing all the time, but every once in a while I think it's good cleanup to do.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9,7 +9,7 @@
   dependencies:
     "@babel/highlight" "^7.10.4"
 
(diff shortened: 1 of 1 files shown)


samselikoff:
Yep totally agree! I wanted to add something to do this regularly actually because you can get in a bad spot with outdated yarn.locks. I've had to manually bisect lockfiles to find problematic deps before... I'll have to tell you about it sometime :)

PR #1002, opened 2021-10-28
Add more useful typescript type for `modelInstance.attrs`

IanVS:
I've found that I sometimes need to get the attributes of a model, but typescript was typing as a generic unknown object.  I _think_ we can type it the same as the properties that are added directly to the model itself, based on the wording of https://miragejs.com/api/classes/model/#attrs.


IanVS:
@zoltan-nz it looks like you added the original TS types.  Does this change seem reasonable to you?

zoltan-nz:
@IanVS Thanks for the update, yep, it looks good to me. Nice improvement. :)

PR #1009, opened 2021-11-30
Route handler return types

pleek91:
This updates the types for `RouteHandler` to allow for any valid JSON response. Adds `number | string | boolean | null` as well as arrays of `object | number | string | boolean | null`.

This is necessary because endpoint can return more than just an object. For instance the project I'm currently working on adding mirage too has an endpoint that returns just a number. Which is totally valid however using the current types without this change causes typescript to emit an error.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -264,7 +264,8 @@ declare module "miragejs/-types" {
   export type AnyRegistry = Registry<AnyModels, AnyFactories>;
 
   type MaybePromise<T> = T | PromiseLike<T>;
-  export type AnyResponse = MaybePromise<ModelInstance | Response | object | number | string | boolean | null>
+  type ValidResponse = object | number | string | boolean | null
+  export type AnyResponse = MaybePromise<ModelInstance | Response | ValidResponse | ValidResponse[]>
 
   /** Represents the type of an instantiated Mirage model.  */
   export type ModelInstance<Data extends {} = {}> = Data & {
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -264,7 +264,7 @@ declare module "miragejs/-types" {
   export type AnyRegistry = Registry<AnyModels, AnyFactories>;
 
   type MaybePromise<T> = T | PromiseLike<T>;
-  type ValidResponse = object | number | string | boolean | null
+  type ValidResponse = Record<PropertyKey, any> | number | string | boolean | null
   export type AnyResponse = MaybePromise<ModelInstance | Response | ValidResponse | ValidResponse[]>
 
   /** Represents the type of an instantiated Mirage model.  */
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -264,8 +264,15 @@ declare module "miragejs/-types" {
   export type AnyRegistry = Registry<AnyModels, AnyFactories>;
 
   type MaybePromise<T> = T | PromiseLike<T>;
-  type ValidResponse = Record<PropertyKey, any> | number | string | boolean | null
-  export type AnyResponse = MaybePromise<ModelInstance | Response | ValidResponse | ValidResponse[]>
+  type ValidResponse =
+    | Record<PropertyKey, any>
+    | number
+    | string
+    | boolean
+    | null;
+  export type AnyResponse = MaybePromise<
+    ModelInstance | Response | ValidResponse | ValidResponse[]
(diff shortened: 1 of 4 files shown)


IanVS:
This makes sense to me.  Would you mind adding some type tests that fail without this change and pass with it?  Thanks!

pleek91:
@IanVS I added some tests though I'm not 100% sure I did them right. Couldn't get them to fail even though vscode showed errors when I didn't have the new types.

types/index.d.ts
```diff
@@ -263,6 +263,10 @@ declare module "miragejs/-types" {
   /** A marker type for easily constraining type parameters that must be shaped like a Registry */
   export type AnyRegistry = Registry<AnyModels, AnyFactories>;
 
+  type MaybePromise<T> = T | PromiseLike<T>;
+  type ValidResponse = object | number | string | boolean | null
```
IanVS, line 267: Instead of `object`, could you please use `Record<PropertyKey, any>`?  `object` in typescript is actually any non-primitive value, which I think is probably not what we want here.

types/tests/server-test.ts
```diff
@@ -22,6 +22,19 @@ export default function config(this: Server): void {
...
+  this.get('/foo', () => true)
+  this.get('/foo', () => null)
+  this.get('/foo', () => ['foo'])
+  this.get('/foo', () => ['foo', 0])
+  this.get<number>('/foo', () => 0)
+  this.get<[number, string]>('/foo', () => [0, 'foo'])
+  
+  this.get<number>('/foo', () => false) // $ExpectError
+  this.get<[number, string]>('/foo', () => ['foo', 0]) // $ExpectError
+
```
IanVS, line 37: I hate to ask this, but would you mind running `yarn prettier:update`?  It looks like the `prettier:check` script does not check `.ts` files in CI, which I'll fix separately.

IanVS (changes requested):
I think the tests are good.  I reverted out your first commit to run the tests without your changes, and they do indeed fail with errors like:

```
ERROR: 26:20  expect  TypeScript@4.0 compile error:
Argument of type '() => number' is not assignable to parameter of type 'RouteHandler<AnyRegistry>'.
  Type 'number' is not assignable to type 'MaybePromise<object | Response | { id?: string | undefined; attrs: {}; modelName: string; save(): void; update<K extends never>(key: K, value: {}[K]): void; update(changes: Partial<{}>): void; destroy(): void; reload(): void; }>'.
```

types/tests/server-test.ts
pleek91: I ran that which made quite a few changes. Please make sure they are correct.

types/index.d.ts
IanVS: I don't think this would be a breaking change, I see it as a bugfix of the types.  Realistically, I wouldn't expect there to be an impact to users, since I'm guessing nobody was relying on an api handler returning a function.  :)

types/tests/server-test.ts
IanVS: Yes I think it's right, but could you back out the changes to `types/tests/collection-test.ts`?  It's making a conflict because I already formatted that file in a separate PR.  Rebasing on latest master might help.  Thanks again.

pleek91: Reverted the changes to `collection-tests.ts` 👍

IanVS (approved):
Looks good, thanks for the PR!

pleek91:
Thanks for your help/input!

pleek91:
How and when would this get released? Just not sure what the process/timeframe is

IanVS:
I'm trying to see if there are any other PRs I can get merged in, and then I'll try cutting a release today.

IanVS:
Turns out I don't have publishing rights to npm, so we need to wait for @samselikoff to do that part.

samselikoff:
Happy to add you & don’t want to block! What’s your npm username

samselikoff:
Done!

IanVS:
Thanks, it looks like I'm added to `ember-cli-mirage`, but not `miragejs`.  I don't use ember, so you may as well remove me from that one.

samselikoff:
Sorry was confused by another thread! Should have it now

IanVS:
@pleek91, 0.1.43 has been released with your changes.  Thanks again, everyone.

PR #1011, opened 2021-12-02
Fix factory types

wagenet:


types/index.d.ts
```diff
@@ -181,7 +181,7 @@ declare module "miragejs/-types" {
   }
 
   type WithFactoryMethods<T> = {
-    [K in keyof T]: T[K] | ((n: number) => T[K]);
+    [K in keyof T]: T[K] | ((this: FlattenFactoryMethods<T>, n: number) => T[K]);
```
jasikpark, line 184: What exactly does this do? 

As I understand it, it takes a factory and translates method return types into basic types, so 

```js
{
    age: number;
    height: () => string;
}
```

becomes

```js
{
    age: number;
    height: string;
}
```

wagenet:
Unfortunately, there's still a lot of messiness around types in Mirage. I'm not sure whether it's still worth merging this so if someone thinks it's good to close, that's ok with me.

IanVS:
Thanks for taking a crack at this, @wagenet.  I'd love to get the types cleaned up, but if I recall correctly, this PR had some failing type tests.  If you're willing to work on them and get them passing, I'll be happy to review.

PR #1019, opened 2022-02-09
Type afterCreate in factories

Istanful:
When creating a factory when using typescript, the afterCreate hook
could not be defined as it is not a key of the model. This commit adds
typing for adding the afterCreate hook.

Unfortunately the exact type of the model is not inferred in this
commit. In order to make the factory fully infer the model, we would
need to create a dependency between the factory type and the model type,
which is something that seems to have been avoided purposefully.

IMHO the any model instance works good enough for now.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -156,7 +156,7 @@ declare module "miragejs" {
 }
 
 declare module "miragejs/-types" {
-  import { Collection, Response } from "miragejs";
+  import { Collection, Response, Server } from "miragejs";
 
   /* A 1:1 relationship between models */
   export class BelongsTo<Name extends string> {
@@ -180,13 +180,15 @@ declare module "miragejs/-types" {
     ): FactoryDefinition<Assign<Data, FlattenFactoryMethods<NewData>>>;
   }
 
-  type WithFactoryMethods<T> = {
-    [K in keyof T]: T[K] | ((n: number) => T[K]);
-  };
+  type WithFactoryMethods<T> =
+    | { afterCreate?: (model: ModelInstance<any>, server: Server) => void }
+    | { [K in keyof Omit<T, "afterCreate">]: T[K] | ((n: number) => T[K]) };
 
   // Extract factory method return values from a factory definition
   type FlattenFactoryMethods<T> = {
-    [K in keyof T]: T[K] extends (n: number) => infer V ? V : T[K];
+    [K in keyof Omit<T, "afterCreate">]: T[K] extends (n: number) => infer V
+      ? V
+      : T[K];
   };
 
   /**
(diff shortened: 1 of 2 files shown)


IanVS:
Thanks for this!  I tested it out in my own project by copying the type file into my node_modules, and while it does allow me to remove my `// @ts-expect-error` comments (\o/), I start to get some other errors unfortunately.  

- I now get a type error on my overall `factories` definition: 
	   ```
	  'factories' implicitly has type 'any' because it does not have a type annotation and is referenced directly or indirectly in its own initializer.
	  ```

- I'm no longer able to access attributes on the factory as I used to.
	image
	vs previously:
	image


- I seem to have lost the types when I use `schema.all()`:
	image
	vs previously: 
	image

I'm not sure what needs to be done to address these, but if you're able to take another crack at it, I'd love to give it another review.  Just send me a ping to be sure I see it.  Thanks again for your efforts!

types/index.d.ts
```diff
@@ -180,13 +180,15 @@ declare module "miragejs/-types" {
     ): FactoryDefinition<Assign<Data, FlattenFactoryMethods<NewData>>>;
   }
 
-  type WithFactoryMethods<T> = {
-    [K in keyof T]: T[K] | ((n: number) => T[K]);
-  };
+  type WithFactoryMethods<T> =
+    | { afterCreate?: (model: ModelInstance<any>, server: Server) => void }
+    | { [K in keyof Omit<T, "afterCreate">]: T[K] | ((n: number) => T[K]) };
```
Nol-go, line 185:
```suggestion
  type WithFactoryMethods<T> = {
    afterCreate?: (model: ModelInstance<T>, server: Server) => void
    [K in keyof T]: T[K] | ((n: number) => T[K])
  };
```

types/index.d.ts
```diff
@@ -180,13 +180,15 @@ declare module "miragejs/-types" {
...
 
   // Extract factory method return values from a factory definition
   type FlattenFactoryMethods<T> = {
-    [K in keyof T]: T[K] extends (n: number) => infer V ? V : T[K];
+    [K in keyof Omit<T, "afterCreate">]: T[K] extends (n: number) => infer V
+      ? V
+      : T[K];
```
Nol-go, line 191:
```suggestion
    [K in keyof T]: T[K] extends (n: number) => infer V ? V : T[K];
```

Nol-go (changes requested):
Try this. I did not test it but I guess it should work

types/index.d.ts
hugo-daclon: Turns out it does not work :cry:

types/index.d.ts
hugo-daclon: Turns out it does not work :cry:

Issue #959, opened 2021-07-20
Typescript - identityManagers property appears to have incorrect type


gfaraj:
This has the identityManagers property with a type of IdentityManager, which does not seem correct:

https://github.com/miragejs/miragejs/blob/master/types/index.d.ts#L299

Is there a workaround for this to build with typescript?

        identityManagers: {
            model1: NumericIdentityManager,
            model2: NumericIdentityManager,
        },

![image](https://user-images.githubusercontent.com/2772908/126379242-ae58300f-6c83-4efb-97cd-80d28d5e398e.png)

gfaraj:
This worked as a work-around:

    identityManagers: {
        model1: NumericIdentityManager,
        model2: NumericIdentityManager,
    } **as any**,

PR #1021, opened 2022-02-16
Fix types for identityManagers

IanVS:
Fixes https://github.com/miragejs/miragejs/issues/959.

@zoltan-nz if you're still around, I'd appreciate your 👀 on this if you've got a chance.  I've validated that it works correctly in my own project, but another set of eyes would be great.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -13,6 +13,8 @@ declare module "miragejs" {
     BelongsTo,
     HasMany,
   } from "miragejs/-types";
+  import IdentityManager from "miragejs/identity-manager";
+  export { IdentityManager };
   export { Server, createServer } from "miragejs/server";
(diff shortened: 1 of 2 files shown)


IanVS:
The failing test seems to be due to dts installing typescript version `4.7.0-dev.20220216` and _calling_ it 4.5.  :(

zoltan-nz:
Hi @IanVS, thanks for improving the type definition file. It looks good to me.

Is there any trick to fix the v17 linting issue? I guess we might have to update the pipeline script.

IanVS:
@zoltan-nz thanks for taking a look!  It's hard to say what's going with ts 4.7.  It could even be a bug in that dev version for all I know.  I wish we could run tests on 4.5 and _maybe_ the 4.6 beta.  Perhaps an update of dtslint would help, but there's an issue to avoid `next` versions that seems to have no movement.

Issue #1026, opened 2022-03-09
Improve types for `params` and `queryParams`


jasikpark:
Currently `params` and `queryParams` are typed as `Record<string,string>` which is true to a certain degree, but it means that Typescript won't let you know that a variable you're accessing might be undefined.

A more accurate typing found in https://github.com/pretenderjs/pretender/blob/65fbf0a608f2e4b4fd79657a77a41f52077074a9/src/pretender.ts#L348: `queryParams?: Record<string, string[] | string | null | undefined>`, etc.

jasikpark:
It looks like `route-recognizer` sets it as `Record<string, unknown>` though my suggestion is probably fine? @IanVS is the one that did the detective work on this though

PR #1027, opened 2022-03-09
Updates typing for `Request.queryParams`

jasikpark:
This uses the type from https://github.com/tildeio/route-recognizer/blob/0a44fd86081ef425071839d3184e72fd2916d6fe/lib/route-recognizer.ts#L432 / https://github.com/tildeio/route-recognizer/blob/0a44fd86081ef425071839d3184e72fd2916d6fe/lib/route-recognizer.ts#L454

I think that https://github.com/tildeio/route-recognizer/blob/0a44fd86081ef425071839d3184e72fd2916d6fe/lib/route-recognizer.ts#L161 might contain the correct type for `params` a la #1026 but that seems to need better verification

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -41,7 +41,7 @@ declare module "miragejs" {
     readonly params: Record<string, string>;
 
     /** Any query parameters associated with the request */
-    readonly queryParams: Record<string, string>;
+    readonly queryParams: Record<string, string[] | string | null | undefined>;
   }
 
   /**
--- a/types/tests/server-test.ts
+++ b/types/tests/server-test.ts
@@ -48,7 +48,7 @@ export default function config(this: Server): void {
     schema.db; // $ExpectType Db
 
     request.params; // $ExpectType Record<string, string>
-    request.queryParams; // $ExpectType Record<string, string>
+    request.queryParams; // $ExpectType Record<string, string | string[] | null | undefined>
     request.requestBody; // $ExpectType string
     request.requestHeaders; // $ExpectType Record<string, string>
     request.url; // $ExpectType string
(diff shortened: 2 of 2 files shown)


IanVS:
@jasikpark could you force-push to this branch to trigger a new run?  I'm curious if it will still fail, and I can't trigger a re-run (maybe it's too old)

jasikpark:
@IanVS mind approving the CI run?

IanVS:
@cah-brian-gantzler how will we handle types when we swap out MSW for pretender?  For instance, in this PR, we can see that the `queryParams` comes from pretender, which uses [`route-recognzier`](https://github.com/tildeio/route-recognizer), which has a type: https://github.com/tildeio/route-recognizer/blob/0a44fd86081ef425071839d3184e72fd2916d6fe/lib/route-recognizer.ts#L433

If one were to swap out MSW, would the types here change, I guess?

IanVS (approved):

jasikpark:
thx for the approval! i don't have merge perms, so merge when you think it's ready to be

cah-brian-gantzler:
@IanVS 
The query params are the params at the end of the url. The url is is the same wether its pretender or msw. I do not believe this would change at all.

jasikpark:
so this can merge then?

cah-brian-gantzler:
I am ok with this being merged

jasikpark:
mind clicking the button for me then? I don't have the permissions for merging my own PR

IanVS:
I will do it in a minute after giving one more quick look.

IanVS:
>The url is is the same wether its pretender or msw. I do not believe this would change at all.

Yes, but the fake `Request` object that is given to the route callbacks comes from pretender today, and presumably from MSW if that gets swapped out.  Happy to run with this, but I think there may be nuances in the types if we allow changing out the mock layer.

Issue #1013, opened 2021-12-22
Decouple from pretender, support MSW others?


andrewluetgers:
Hi I am coming to this project from using MSW and realized organically that mocking the api is the best solution vs one off snapshots. This is clearly the project for doing such things, very nice! I however really appreciate how MSW works and has integrated nicely with my testing and Storybook development workflows. How hard would it be for me to get this working in MSW instead of leaning on pretender? Any guidance or previous attempts at this would be greatly appreciated. 

I think going forward this project would grow even faster if this was configurable, mock with pretender, or MSW or whatever. Case in point, Kent C. Dodds seems to let the issue of how, (technically) you mock requests, blind him to the benefits of why domain modeling and mocking an api with mirage.js is superior to other methods. https://kentcdodds.com/calls/01/13/mirage-js-over-msw Making this issue go away could mean many more people using mirage.js

IanVS:
I personally would be very on board with this, but I have no idea how tightly coupled mirage is to pretender.  From what I can tell, @samselikoff wouldn't be opposed to it, and may even have some thoughts on how it could be done.  Sam, a bit of guidance here could go a long way towards moving this project forward, IMHO.

jneurock:
I'm not aware of previous attempts but I have thought about the same thing and would love to figure out how much effort it would be. Maybe something nice to work on over the holiday break.

samselikoff:
I haven't been personally maintaining Mirage for a while now but happy to give any advice. Maybe a call in the future would be helpful if folks were really willing to take on the work.

Here's some history:

- https://github.com/miragejs/miragejs/issues/210
- https://github.com/miragejs/miragejs/pull/699/files
- https://github.com/miragejs/miragejs/pull/320

I'll have to ask @ryanto if there was more code on our machines somewhere but the general idea was to make an async  `server.handle(request)` method that would return the response as just a JS response object (req/res signature undecided, ideally something based on browser APIs or copy from express). Then use something of an adapter pattern so people can hook that into pretender (backwards-compat), or use it within an express route handler (to run it in node), or in msw (or whatever over "interceptor" layer they prefer).

[...]

PR #1036, opened 2022-04-23
Extract pretender code to its own module for future move to another package

cah-briangantzler:
This PR moves all the pretender related code to the file `mock-server/pretender-config` and conditions the creation of pretender on the presence of an additional server option `interceptor`.

The goal of this PR was to be fully backward compatible. It opens up the ability to replace the interceptor with a different interceptor or allow for no interceptor at all.

First step towards https://github.com/miragejs/miragejs/issues/1013

### DIFF
--- a/__tests__/external/shared/orm/belongs-to/1-basic/association-set-test.js
+++ b/__tests__/external/shared/orm/belongs-to/1-basic/association-set-test.js
@@ -1,6 +1,6 @@
 import Helper, { states } from "./_helper";
 
-describe("External | Shared | ORM | Belongs To | Basic | association #set", () => {
+describe("External | Shared | ORM | Belongs To | Basic | association #set", function () {
   let helper;
   beforeEach(() => {
     helper = new Helper();
--- a/__tests__/internal/move-after-handle-request/get-full-path-test.js
+++ b/__tests__/internal/move-after-handle-request/get-full-path-test.js
@@ -17,21 +17,27 @@ describe("Integration | Server | Get full path", function () {
     expect.assertions(1);
     server.namespace = "/api";
 
-    expect(server.interceptor._getFullPath("/contacts")).toEqual("/api/contacts");
+    expect(server.interceptor._getFullPath("/contacts")).toEqual(
+      "/api/contacts"
+    );
   });
 
   test("it works with a configured namespace with a trailing slash", () => {
     expect.assertions(1);
     server.namespace = "api/";
 
-    expect(server.interceptor._getFullPath("/contacts")).toEqual("/api/contacts");
+    expect(server.interceptor._getFullPath("/contacts")).toEqual(
+      "/api/contacts"
+    );
   });
 
   test("it works with a configured namespace without a leading slash", () => {
     expect.assertions(1);
     server.namespace = "api";
 
-    expect(server.interceptor._getFullPath("/contacts")).toEqual("/api/contacts");
+    expect(server.interceptor._getFullPath("/contacts")).toEqual(
+      "/api/contacts"
+    );
   });
 
   test("it works with a configured namespace is an empty string", () => {
--- a/__tests__/internal/move-after-handle-request/get-full-path-test.js
+++ b/__tests__/internal/move-after-handle-request/get-full-path-test.js
@@ -17,35 +17,35 @@ describe("Integration | Server | Get full path", function () {
     expect.assertions(1);
     server.namespace = "/api";
 
-    expect(server._getFullPath("/contacts")).toEqual("/api/contacts");
+    expect(server.interceptor._getFullPath("/contacts")).toEqual("/api/contacts");
   });
 
   test("it works with a configured namespace with a trailing slash", () => {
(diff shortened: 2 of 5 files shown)


IanVS:
Thanks for this work, @cah-briangantzler!  I'm not terribly familiar with the pretender/mirage codebase (I mostly hopped in as a maintainer to keep dependencies and types up to date), but I really appreciate the work you're doing here, and I think it's a move in the right direction.  I'd like to ask @samselikoff if he has any concerns with the process you've outlined in https://github.com/miragejs/miragejs/issues/1037, and if not, I say let's go for it.  I'll give this a review as soon as I can, but I'm a bit swamped at the moment, so please ping me later this week if you haven't heard anything back.

cah-briangantzler:
Thanks for the feedback. Yea, there definitely room for improvement but I wanted to open the ability to experiment (I already have MSW mostly working) without being a breaking change and this does that. There will be breaking in the future, trying to keep that to a minimum

IanVS (approved):
Tests are passing and the code looks reasonable to me.  Let's do this!

IanVS:
🎉 This has been published in https://github.com/miragejs/miragejs/releases/tag/v0.1.44.  I know it might not have _needed_ a release, but I figure it's best to take it slow, and get feedback from users if there are accidental breaking changes.  Thanks again for your efforts, @cah-briangantzler!

kpfefferle:
@cah-briangantzler 

We're getting an error in one of our applications when adopting this update (but not in our other application - which is odd). Both applications are using `miragejs` via `ember-cli-mirage@3.0.0-alpha.3`. It works fine with `miragejs@0.1.43` but the following errors blow up the test suite with `miragejs@0.1.44`:

```
Mirage: Error: You cannot modify Pretender's request tracking once the server is created
                at new MirageError (webpack://__ember_auto_import__/../node_modules/miragejs/dist/mirage-esm.js?:181:47)
                at assert (webpack://__ember_auto_import__/../node_modules/miragejs/dist/mirage-esm.js?:177:7)
                at PretenderConfig.config (webpack://__ember_auto_import__/../node_modules/miragejs/dist/mirage-esm.js?:2498:1166)
                at PretenderConfig.create (webpack://__ember_auto_import__/../node_modules/miragejs/dist/mirage-esm.js?:2498:103)
                at Server.config (webpack://__ember_auto_import__/../node_modules/miragejs/dist/mirage-esm.js?:2695:496)
                at new Server (webpack://__ember_auto_import__/../node_modules/miragejs/dist/mirage-esm.js?:2683:163)
                at createServer (webpack://__ember_auto_import__/../node_modules/miragejs/dist/mirage-esm.js?:2656:42)
```

```
{"type":"warn","text":"You created a second Pretender instance while there was already one running. Running two Pretender servers at once will lead to unexpected results and will be removed entirely in a future major version.Please call .shutdown() on your instances when you no longer need them to respond."}
```

cah-briangantzler:
That is odd. The code as moved around should be doing the same thing. 

The main thing here is that it thinks you are creating a second pretender service which was never allowed. The pretender.shutdown is typically called via the afterHook from setupMirage. 

Is it possible a test is starting before another test has completed?

Does this happen on the first test? or after several tests have run? If always the first test, I would look to see if autostart is somehow starting a server. (Hmm, think i removed autostart). 

I would put a breakpoint on the pretender start https://github.com/miragejs/miragejs/blob/master/lib/mock-server/pretender-config.js#L238 and stop  https://github.com/miragejs/miragejs/blob/master/lib/mock-server/pretender-config.js#L529 to see where the second start is happening. 

The assert for the not changing the tracking request is here https://github.com/miragejs/miragejs/blob/master/lib/mock-server/pretender-config.js#L131. Thats original code and I dont actually understand what its trying to do. There is a trackRequests option you can do, but I have never done it. Does you second app set the trackRequests property and your first app doesnt? It may be that the track requests isnt handled properly in the moved code.

IanVS:
Thanks for letting us know, @kpfefferle.  If you're able to create a minimal reproduction of the issue, I think that would be the most helpful way we can dig into it, aside from the troubleshooting tips that @cah-briangantzler provided.

the-kwisatz-haderach:
I seem to run into the same issue as @kpfefferle after updating to v0.1.44. Unfortunately I don't have time to create a minimal reproduction at the moment. Maybe during the weekend. I've tried running a single test, creating a single server, without setting the `trackRequests` option and still getting this issue.

IanVS:
@cah-briangantzler there have been several reports now of this issue.  Do you have any time to look into it so we can fail forward, or should I plan to revert this PR until we can figure out what's going on?

cah-briangantzler:
The track request code is pretty small. Yea, Ill add track requests to my config and see what happens

the-kwisatz-haderach:
I can at least confirm that it works for me now after upgrading to `0.1.45`! 🥳

cah-briangantzler:
Thats great to hear, hope @kpfefferle will get a chance to try it soon.

Issue #1043, opened 2022-05-27
I simply can't use/config trackRequests anymore and I'm not sure why


richardnikolas:
I had to recreate my test environment and I was using `miragejs` just fine, tracking all requests and asserting what I needed. 
But, on this new test environment (using React), I keep getting the same freaking error. I'm using version `0.1.44`.

**Mirage: Error: You cannot modify Pretender's request tracking once the server is created**

My config is like this now: 

```javascript
export function makeServer() {
    return createServer({
        environment: "test",
        trackRequests: true,
        serializers: {
            application: ApplicationSerializer
        },
```

I've already tried removing the `environment`, removing the `trackRequests` as well (hoping it would track if it was a test environment), etc. But nothing works. 

And after removing the trackRequests from the configuration, I saw this on the `mirageServer` log:

![image](https://user-images.githubusercontent.com/20261986/170784892-4ec4acc8-fc8c-43f1-a7d6-743f43a5e79b.png)

Anyone with this problem as well? Please help :(

IanVS:
@richardnikolas can you try using version 0.1.43?  It seems that https://github.com/miragejs/miragejs/pull/1036 may have caused some problems, and we might need to revert it.

richardnikolas:
@IanVS I tried using 0.1.42 and didn't work as well (because that's the one I was using on my old project). But yeah, I can try 0.1.43 as well.

IanVS:
>I tried using 0.1.42 and didn't work as well

Were you getting the same error, or a different one?

richardnikolas:
> Were you getting the same error, or a different one?

I was getting the same exact error.

richardnikolas:
Sorry, @IanVS . I was using `^0.1.42` on my package.json 😑, so it was installing the newer version anyway. 
But now the tracking is working again with the _real_ `0.1.42`.

cah-briangantzler:
Thanks for reporting. Have released a fix. I turned on trackedRequests in my app and all tests ran. I debugged a few of the tests and verified there were requests captured in `this.server.pretender.handledRequests`

IanVS:
The fix hasn't been released just yet, but I'll try to get it out ASAP.

cah-briangantzler:
Thanks so much for the release. Although I dont use the trackRequests, I have updated my app to 0.1.45 and all tests ran successfully.

PR #1045, opened 2022-06-01
Removed assert about changing Pretender config. Ensured trackRequest was set properly

cah-briangantzler:
Fixed the assert that informed you could not change the trackRequests after pretender is created. 

After moving things around, the value of tracked Request was being sent into pretender via a function. The value in the function was being set too late. Since the value was already known, just simplified as putting it behind a function served no purpose.

Fixes https://github.com/miragejs/miragejs/issues/1043

### DIFF
--- a/lib/mock-server/pretender-config.js
+++ b/lib/mock-server/pretender-config.js
@@ -131,7 +131,8 @@ export default class PretenderConfig {
     }
 
     let didOverridePretenderConfig =
-        config.trackRequests !== undefined && config.trackRequests !== this.trackRequests;
+      config.trackRequests !== undefined &&
+      config.trackRequests !== this.trackRequests;
     assert(
       !didOverridePretenderConfig,
       "You cannot modify Pretender's request tracking once the server is created"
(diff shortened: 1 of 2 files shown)


cah-briangantzler:
Seems there is a test that actually checks that the assert is thrown.  Rather than remove the assert, will see if code can be restructured to still make the assert possible

cah-briangantzler:
I have made these same changes in `mirage-pretender`. Will release a version of that once this is merged and confirmed to fix the issue.

lib/mock-server/pretender-config.js
```diff
@@ -129,7 +131,8 @@ export default class PretenderConfig {
     }
 
     let didOverridePretenderConfig =
-      config.trackRequests !== undefined && this.pretender;
+      config.trackRequests !== undefined &&
+      config.trackRequests !== this.trackRequests;
```
IanVS, line 135: So, instead of checking whether there's a `pretender` instance, we look to see whether the `trackRequests` has changed.  This makes sense, although the name of `didOverridePretenderConfig` feels overly broad for what's actually being checked.  But, I think that's fine.

IanVS (approved):
LGTM, thanks for hopping into this one.

lib/mock-server/pretender-config.js
cah-briangantzler: Its a local var, I was just changing the minimum

Issue #1049, opened 2022-07-12
Route timing property no longer works after upgrading from 0.1.43 to 0.1.45


bewards:
Individual (per) route timing is no longer respected after upgrading.

**System info**
Node v16

bewards:
```typescript
import { createServer, Factory, Model, Response, RestSerializer } from "miragejs";
import { faker } from "@faker-js/faker";

const ApplicationSerializer = RestSerializer.extend({
    embed: true
});

createServer({
    environment: "development",
    serializers: {
        application: ApplicationSerializer
    },
    routes() {
        this.post('/api/Account/ForgotUsername', () => {
            const flag = faker.datatype.boolean();
            const response: ForgotUserResponse = { isSuccess: flag };
            if (!flag) {
                response.errorMessage = "<p>There was a problem submitting your request. Please reload the page and try again or call WIN's Customer Care.</p>";
            }

            return response;
        }, { timing: 1000 });
    }
});
```

cah-briangantzler:
Yep, I see where it is broken. Will try to get a fix out soon

https://github.com/miragejs/miragejs/blob/v0.1.45/lib/mock-server/pretender-config.js#L112 should not be using the fifth argument instead of config.

NikJ87:
@cah-briangantzler 
Apologies, just want to know when are we expecting fix for it. Meanwhile I'm failing back to 43.

Thanks Nik

cah-briangantzler:
Taking a little extra time to see what I can do about reconciling responsibilities without introducing more breaking change.  

Work has also been busy

PR #1051, opened 2022-07-29
Fixed timing being incorrectly passed to pretender

cah-briangantzler:
Fixes: https://github.com/miragejs/miragejs/issues/1049

This required a change to the contract for pretender being extracted and updating the extracted pretender. Since we have not removed the internal pretender code yet, this version will work. Should anyone be using the extracted version of pretender at https://github.com/bgantzler/mirage-pretender, that will break until I update that repo as well. Since it hasnt been officially released (waiting for this repo to settle down and shake out bugs, this issue being a prime example), I dont think it will be an issue. 

There were no tests testing this sort of thing to update and I didnt write any as Im not sure exactly how to test it given its the communication between mirage and pretender and kinda happens behind the scenes, which might have been why no tests exist. But since this is a production bug I hope it can be merged anyway. I did install this version in my own app and apply some timings to some of the mirage routes and the tests reacted accordingly.

### DIFF
--- a/lib/mock-server/pretender-config.js
+++ b/lib/mock-server/pretender-config.js
@@ -2,6 +2,7 @@ import "@miragejs/pretender-node-polyfill/before";
 import Pretender from "pretender";
 import "@miragejs/pretender-node-polyfill/after";
 import assert from "../assert";
+import assign from "lodash.assign";
 
 /**
   Mirage Interceptor Class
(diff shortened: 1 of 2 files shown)


lib/mock-server/pretender-config.js
```diff
@@ -47,11 +48,65 @@ const defaultPassthroughs = [
...
+const defaultRouteOptions = {
+  coalesce: false,
+  timing: undefined,
+};
+
+/**
+ * Determine if the object contains a valid option.
+ *
+ * @method isOption
+ * @param {Object} option An object with one option value pair.
```
IanVS, line 60: >An object with one option value pair

Does this mean that only one option can be set at a time?  From reading the code, it looks like it's checking that there's _at least_ one option key being used, but there could be other keys as well.

lib/mock-server/pretender-config.js
```diff
@@ -47,11 +48,65 @@ const defaultPassthroughs = [
...
  @hide
  */
 export { defaultPassthroughs };
 
+/**
+ * Extract arguments for a route.
+ *
+ * @method extractRouteArguments
+ * @param {Array} args Of the form [options], [object, code], [function, code]
+ * [shorthand, options], [shorthand, code, options]
```
IanVS, line 90: This seems like a pretty funky format.  Maybe an example here would help make this more clear.  That's just a nice-to-have, though.

Edit, oh I see, you just moved this from elsewhere.  Ignore these couple of comments if you like.

IanVS (approved):
Looks reasonable to me.

lib/mock-server/pretender-config.js
cah-briangantzler: It was the code that was already there supporting this, I just moved it into the extracted version of the pretender code.

lib/mock-server/pretender-config.js
cah-briangantzler: Yes, this code was just moved around from the original code

Issue #1052, opened 2022-08-02
Typescript error on afterCreate in factory


tuzmusic:
I have the following code for my factory, omitting some properties for brevity. 

```typescript
export const propsPlayerDetailFactory = Factory.extend({
  sportId: () => getFakeSportId(),
  playerId: i => `${i}`,
  playerName: () => faker.name.findName(),
  playerProps: [] as PropAdminActiveGridRecord[],
  afterCreate(player, server) {
    player.update({ playerProps: server.createList('playerProp', 15) });
  },
});
```

`afterCreate` gives a Typescript error, but does work properly (when the error is ignored).

> TS2322: Type '(player: any, server: any) => void' is not assignable to type 'void | ((n: number) => void)'.   Type '(player: any, server: any) => void' is not assignable to type '(n: number) => void'.  propsPlayer.factory.ts(37, 3): 
>
>The expected type comes from property 'afterCreate' which is declared here on type 'WithFactoryMethods<{ sportId: string; playerId: string; playerName: string; playerProps: PropAdminActiveGridRecord[]; afterCreate: void; }>'

I have type definitions in various other places to define the `AppRegistry`, but since those defs use `propsPlayerDetailFactory ` and not the other way around, it seems like `Factory.extend` is the only type that's relevant here.

Can we get type support for factory hooks?

cah-briangantzler:
Definitely welcome a PR to add typescript support. I have not gotten into typescript yet, so would not be able to provide at this time.

IanVS:
FWIW, there have been a few attempts at it so far:

https://github.com/miragejs/miragejs/pull/525
https://github.com/miragejs/miragejs/pull/797
https://github.com/miragejs/miragejs/pull/1019

Maybe they can provide some clues as to how to go about it.

Apetree100122:
___export const propsPlayerDetailFactory = Factory.extend({ __sportId: () => get FakeSportId (__),  playerId:_ i = _>`${   i  }`, playerName: (__) = > faker.name.find Name), playerProps: 
[__] as Prop_Admin_Active_Grid_Record[__],after_Create__(player,server__) __{  __player.update({ playerProps: __server.create__List('__player__Prop', 15__)__ }__)__;},__}__);

Apetree100122:
after_Create 
gives a Type__script error, 
but does work properly(
when the error is ignored__).
    TS
__2322__: Type '(__player: any__,server: any) __
> void' 
[...]

PR #1054, opened 2022-08-07
Add Typescript support for afterCreate

tuzmusic:
To address #1052. Includes a test in `factory-test.ts`.

By the way, in that test file, `// $ExpectError` still shows as a TS error, although that might just be for me in Webstorm. `// @ts-expect-error`, placed above the line, shows no red wavy line. Although I don't know if there's a similar equivalent for `$ExpectType`. @zoltan-nz is there anywhere that these "tests" are run or anything?

I would be happy to swap those error assertions as part of this PR, or alternatively in a separate PR, if you'd like. Let me know!

(🥳 my first OSS PR!)

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -158,7 +158,7 @@ declare module "miragejs" {
 }
 
 declare module "miragejs/-types" {
-  import { Collection, Response } from "miragejs";
+  import { Collection, Response, Server } from "miragejs";
 
   /* A 1:1 relationship between models */
   export class BelongsTo<Name extends string> {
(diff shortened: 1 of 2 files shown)


IanVS:
Hi, thanks for this PR!  To answer your question, we do test the typescript definitions in CI: https://github.com/miragejs/miragejs/runs/7725993547?check_suite_focus=true#step:6:1, and it looks like they are passing in this PR.  However, there's a formatting error that is failing.  You can run `yarn prettier:update` and check in the results to fix that one.  I'll give this a better review and try it out in my own project shortly.

IanVS:
It looks like our prettier commands are not working the way we want.  Sorry to ask you this, but could you run `yarn prettier --write types/tests/factory-test.ts` and check that in?  I'll fix the package.json script in another PR.  Or, if you're having trouble, I'll check out your branch and push the fix in a day or two.

cah-briangantzler:
Forgetting to run prettier always bytes me on PRs. For my own repos I put husky in place to run prettier on commits just to avoid that

IanVS:
@tuzmusic it's prettier that's failing again.  As I mentioned above, it seems that our `update` script does not actually format typescript files, so it missed the test.  Running `yarn prettier --write types/tests/factory-test.ts` should take care of it.

IanVS:
That's strange.  OK, don't worry about it, I'll look into it.  Thanks again for putting together the PR.

IanVS:
I'm testing this out in my own app and getting a lot of type errors.  I'll try to figure out what's going on and report back.

PR #1059, opened 2022-09-22
Add resource function type definition

morganmspencer:
Adds the resource function type definition to the Server class

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -322,12 +322,14 @@ declare module "miragejs/server" {
     /** A number of ms to artificially delay responses to this route. */
     timing?: number;
   }
+
+  type ShorthandOptions = 'index' | 'show' | 'create' | 'update' | 'delete'
   
   export interface ResourceOptions {
     /** Whitelist of shorthand options */
-    only?: string[];
+    only?: ShorthandOptions[];
     /** Exclude list of shorthand options */
-    except?: string[];
+    except?: ShorthandOptions[];
     /** Shorthand route path */
     path?: string;
   }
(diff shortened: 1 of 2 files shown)


cah-brian-gantzler:
Hoping @IanVS can look at this. I am not using typescript yet

types/index.d.ts
```diff
@@ -322,6 +322,15 @@ declare module "miragejs/server" {
     /** A number of ms to artificially delay responses to this route. */
     timing?: number;
   }
+  
+  export interface ResourceOptions {
+    /** Whitelist of shorthand options */
+    only?: string[];
+    /** Exclude list of shorthand options */
+    except?: string[];
```
IanVS, line 330: We can be a bit more helpful here, and use a union of the allowed values.

The values allowed are `'index' | 'show' | 'create' | 'update' | 'delete'`

types/index.d.ts
```diff
@@ -452,6 +461,9 @@ declare module "miragejs/server" {
       handler?: RouteHandler<Registry, Response>,
       options?: HandlerOptions
     ): void;
+    
+    /** Define multiple shorthands for a given resource */
+    resource(model: string, options?: ResourceOptions): void;
```
IanVS, line 466: Let's restrict the modelName to valid values, like what's done in `createList` above.  

```suggestion
    resource<K extends keyof Registry>(modelName: K, options?: ResourceOptions): void;
```

IanVS (changes requested):
Thanks for the PR!  I made a few suggestions to help narrow down the typing a bit, and would you mind adding a test in https://github.com/miragejs/miragejs/blob/master/types/tests/server-test.ts as well?

IanVS (approved):

PR #1061, opened 2022-09-30
Align passthrough type declaration with implementation

brzosthub:
It seems like passthrough type declaration is incorrect. It suggest that you can pass in array of for e.q strings. When we look on implementation passthrough is implemented as ```passthrough(...paths)```. Fix type. Using ```passthrough(['url1', 'url2'])``` will not work.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,4 +1,4 @@
-// Minimum TypeScript Version: 3.7
+// Minimum TypeScript Version: 4.2
 
 /*
  * Inspired by Dan Freeman
@@ -298,7 +298,7 @@ declare module "miragejs/-types" {
 }
 
 declare module "miragejs/server" {
-  import { Request, Response, Registry as MirageRegistry } from "miragejs";
+  import { Request, Registry as MirageRegistry } from "miragejs";
   import {
     AnyRegistry,
     AnyModels,
@@ -306,11 +306,26 @@ declare module "miragejs/server" {
     AnyResponse,
     Instantiate,
   } from "miragejs/-types";
-  import { ModelInstance } from "miragejs/-types";
+
   import Db from "miragejs/db";
   import IdentityManager from "miragejs/identity-manager";
   import Schema from "miragejs/orm/schema";
   import PretenderServer from "pretender";
+
+  /**
+   * Possible HTTP verbs
+   * @see https://github.com/pretenderjs/pretender/blob/master/index.d.ts#L13
+   **/
+  type HTTPVerb =
+    | "get"
+    | "put"
+    | "post"
+    | "patch"
+    | "delete"
+    | "options"
+    | "head";
+  type PassthroughUrl = ((request: Request) => any) | string;
+  type PassthroughVerbs = HTTPVerb[];
 
   /** A callback that will be invoked when a given Mirage route is hit. */
   export type RouteHandler<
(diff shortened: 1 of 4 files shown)


IanVS:
Thanks for this.  After reading the implementation, I think you're correct, multiple functions could be provided.  However, I think we also need to account for the last element passed to also be an array (of verbs), as shown here: https://github.com/miragejs/miragejs/blob/master/__tests__/external/browser-only/passthrough-test.js#L51.  Ideally, we should try locking that down such that only [valid verbs](https://github.com/miragejs/miragejs/blob/f68581a1b56c4d3e5efcee94641fa761fd3add60/lib/mock-server/pretender-config.js#L450) can be used in an array in the last argument.  But I'm afraid my TS skills aren't strong enough to suggest how to actually do that.

types/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Minimum TypeScript Version: 3.7
+// Minimum TypeScript Version: 4.2
```
brzosthub, line 1: We are using variadic tuples with spread operator in the middle, typescript 4.2 is minimum for that.

types/index.d.ts
```diff
@@ -298,20 +298,35 @@ declare module "miragejs/-types" {
...
+  import { Request, Registry as MirageRegistry } from "miragejs";
   import {
     AnyRegistry,
     AnyModels,
     AnyFactories,
     AnyResponse,
     Instantiate,
   } from "miragejs/-types";
-  import { ModelInstance } from "miragejs/-types";
+
```
brzosthub, line 309: Those are jus removals of unused things...

types/index.d.ts
```diff
@@ -454,7 +469,10 @@ declare module "miragejs/server" {
     ): void;
 
     /** Pass through one or more URLs to make real requests. */
-    passthrough(urls?: ((request: Request) => any) | string | string[]): void;
+    passthrough(...urls: PassthroughUrl[]): void;
+    passthrough(
+      ...args: [PassthroughUrl, ...PassthroughUrl[], PassthroughVerbs]
+    ): void;
```
brzosthub, line 475: Here we have new types for pass through, its either just urls, or at least one url followed by array of verbs.

types/tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "target": "es6",
     "lib": ["es6", "dom"],
     "module": "ES6",
+    "types": [],
```
brzosthub, line 6: We are removing node types from types checks if we do not do it we will get:

`../node_modules/@types/node/globals.d.ts(72,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'AbortSignal' must be of type '{ new (): AbortSignal; prototype: AbortSignal; timeout(milliseconds: number): AbortSignal; }', but here has type '{ new (): AbortSignal; prototype: AbortSignal; }'.`

In general it does not make sense to have types: ['node'] (default) because it is browser running project.

brzosthub:
@IanVS I've updated PR:
- types are updated to follow your suggestions
- types test are updated
- build is fixed
- added some comments to explain what and why

types/tests/server-test.ts
```diff
@@ -36,7 +37,14 @@ export default function config(this: Server): void {
   this.get<[number, string]>("/foo", () => ["foo", 0]); // $ExpectError
 
   this.passthrough("/_coverage/upload"); // $ExpectType void
+  this.passthrough("/_coverage/upload_a", "/_coverage/upload_b"); // $ExpectType void
+  this.passthrough(["/_coverage/upload"]); // $ExpectError
   this.passthrough((request) => request.queryParams.skipMirage); // $ExpectType void
+  this.passthrough("/_coverage/upload", ["get"]); // $ExpectType void
+  // prettier-ignore
+  this.passthrough("/_coverage/upload", (request) => request.queryParams.skipMirage); // $ExpectType void
+  // prettier-ignore
```
brzosthub, line 46: prettier-ignore is required otherwise we will get line break and dtslint will not understand end line annotation

brzosthub:
@IanVS can that one be reviewed and merged?

IanVS:
Thanks for the reminder, I'll take a look as soon as I can, hopefully today.

IanVS (approved):
This looks good to me, thanks for fixing our failing tests.

PR #1064, opened 2022-10-25
Fix includes query params

mansona:
Howdy folks 👋 

I came across a problem with ember-cli-mirage crashing on `queryParamIncludes.split(",")` code that I've changed in this PR. When debugging it it looks like `request.queryParams.include` is actually an array and not a string as the code expects 🤔 

Looking at the test it seems like we're testing a mock object that looks like a request and my suspicion is that when you're doing this for real with a FakeRequest (from faker) the query param has already been converted into a real array instead of a comma separated string. To be sure I thought the best approach was to just make sure that it's not an array before trying to call split on it, and I duplicated the test that covers this and passed in an array in the mock object.

Let me know if you would like me to change anything else before merging 👍


cah-brian-gantzler:
You are correct. Big file, scrolling up and down I crossed paths.

mansona:
lol no worries, I completely understand 😂 I was looking at it on my phone yesterday trying to understand what you were saying and I was like "nope! file too big 🙈 will reply at my computer tomorrow" 😂

IanVS:
Sure thing, I just wanted to get CI to run with the latest changes, since some of the previous failures have been fixed on `master`

mansona:
Hey folks, I'm just checking if anyone had a chance to look at this yet? I tried to be as careful (and as neat 😂 ) as possible so that existing functionality hasn't changed at all while the bug we experienced in our codebase is covered.

Are there any blockers for getting this merged and released? I can rework it if anyone would prefer 👍

IanVS:
@cah-brian-gantzler mind if I leave this one in your hands?

cah-brian-gantzler:
Sure, I already looked at it when I commented earlier, will check again once I get more time.

cah-brian-gantzler (approved):

mansona:
Hey thanks @cah-brian-gantzler for merging this 🎉 

Is there any chance we could get a release? Usually I would just target the git master while I was waiting but because of that build step it doesn't work 😞

IanVS:
Sure, we've got a few changes stacked up, I'll try to get a release out today.

cah-brian-gantzler:
Thanks. I have been waiting on a couple changes to be released as well. Have to make some changes to mirage-pretender and mirage-msw.   

After this release, should be able to make the PR to remove pretender from mirageJS and depend on the mirage-pretender repo

PR #1066, opened 2022-11-12
Improve ModelInstance<Model>.update

lanhhv84:
- Given:
```
type Model =  {
    items: HasMany<Item>
}
```

- Current behavior:
```
ModelInstance<Model>.update({items: [item1, item2]}); // $ExpectError
ModelInstance<Model>.update({items: Collection<Item>});
```

- New behavior:

```
ModelInstance<Model>.update({items: [item1, item2]}); 
ModelInstance<Model>.update({items: Collection<Item>}); 
````

- Motivation: We should be able to use readily available/simple data types to update our models, instead of having to create a new `Collection` first before we can update.

### DIFF
--- a/lib/orm/model.js
+++ b/lib/orm/model.js
@@ -143,17 +143,21 @@ class Model {
     Updates the record in the db.
 
     ```js
+    let author = authors.find(1);
+    let followers = users.find([1, 2]);
     let post = blogPosts.find(1);
     post.update('title', 'Hipster ipsum'); // the db was updated
     post.update({
       title: 'Lorem ipsum',
       created_at: 'before it was cool'
     });
+    post.update({ author });
+    post.update({ followers });
     ```
 
     @method update
     @param {String} key
-    @param {String} val
+    @param {any} val
     @return this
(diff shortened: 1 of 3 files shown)


IanVS:
I guess this would be a breaking change for anyone providing items as a collection?

lanhhv84:
Good point! I will update the code to suport both Array and Collection. Thanks.

lanhhv84:
Thanks for your feedback. I updated the code to remove breaking changes.

IanVS (commented):
This looks good to me, but it seems like maybe the documentation is incorrect.  It seems to indicate that the second argument to `update` should only be a `string`.  Or am I understanding it incorrectly? [model.update](https://miragejs.com/api/classes/model/#update) I'm guessing the docs are just wrong, ya?  If so, do you think you could find and fix that as well?  It should be here in this repo.

lanhhv84:
Great catch! The documentation for that function should be updated with the code now.

IanVS:
Thanks for the improvements!

PR #1067, opened 2022-11-13
Setup Typescipt and make Collection Array-like

lanhhv84:
## Motivation
- Given these models
```ts
type Address {}

type Car {}

type Person {
    address: Address; // One-To-One
    cars: Car[]; // One-To-Many
}
```

- When we write a component, it is natual that we just use the native JS data types (Array, string, number,...). I think It could better that the data returned by MirageJs can act like native data types, so that we can pass these data directly to the component.

```ts
const person = server.create(Models.Person);
render(<Component person={person}/>);
/*
const Compoent = ({person: Person}) => {
    return <>
    ...
   {
      person.cars.map((car) => ...)
   }
   ...
   </>;
}
*/
```
## Changes
- I create a new Collection type, which extends Array type and have all the existing functionalities.

- Since it is a rewrite, I think it is also a good time to start to write the code in Typescript (?)


IanVS:
Hi @lanhhv84, thanks for this, it looks like you've already put a lot of work into it.

I see a move to typescript as a bit risky of an endeavor, especially since the original creator of Mirage is no longer involved.  Changing the functionality at the same time as migrating to typescript is even more tricky.

What would you think about taking this one step at a time?  If you're willing to work on a typescript re-write, maybe we can start a long-running branch for that, which we could tackle a small piece at a time, merging PRs into it bit by bit.  Then we can confirm that the types generated by TSC match those we have manually written so far, to be sure there are no unintentional breaking changes.

Then, after that, we could explore tweaking the types as you've done here, which might be easier to review and roll out in isolation from the typescript refactor.  

What do you think about that, would that be something you'd be interested in working on, or is that too large of an undertaking?  

@samselikoff, I know you're not involved anymore, but I thought I'd at least mention you here as this is a bit bigger of a move, if we're talking about migrating the codebase to TypeScript (I'm generally in favor, if we can do it carefully and safely).  I think due to some of the dynamic nature of how Mirage works, it might also be tricky to do.

lanhhv84:
Hi Ivan, thank you for your suggestion.

I don’t mind rewriting the library in Typescript. I think it should be a feasible task as I have been a long-time user of this library with experience working with most data types.

Regarding whether we should merge tiny changes one at a time into master or a new long-running branch, I don’t have a preference as I am pretty new to the OSS world. I might rely on you for this decision :)

IanVS:
>I don’t mind rewriting the library in Typescript. I think it should be a feasible task as I have been a long-time user of this library with experience working with most data types.

🎉 That's great news!  You probably have more experience with it than I do, then, since I do not use Mirage to its full capacity.  

@cah-brian-gantzler, what do you think?  You've mentioned that you haven't used typescript, would it be a problem for you if this repo were converted to typescript?  I think you'll find that it will become more understandable overall.  And I don't think either of us are planning to add new features or changes that would require any TS wizardry on our part, so I'm inclined to go ahead with it.

As for the mechanics of it if we do, I'd suggest creating a feature branch and merging PRs into it.  They don't necessarily need to be "tiny", but some manageable / reviewable chunk at a time would be good.  We can release alpha versions as we go, in case we want to try it out to be sure things haven't broken as we go.  I'd suggest starting with whatever part of the code you expect to be most difficult, to get that out of the way and find out if there will be roadblocks early.

Thanks for volunteering for this, @lanhhv84.  Managing the TS types as disconnected from the code has been a challenge, as you can tell from all the PRs around typescript improvements, many of which still can't be merged because they're not working correctly.  Between converting to TS and @cah-brian-gantzler's work to decouple mirage from pretender, I think there's a lot of opportunity to breathe new life into this project.

cah-brian-gantzler:
While I havent had as much exposure to typescript as I would like, I have no problem with this being updated to typescript. Ember is typescript so I have to read it when debugging anyway :)

samselikoff:
I think you can go for it! I would definitely change some of the APIs to be both simpler and thus more easy to type but I know there's a lot of Mirage code out there so the backwards-incompat changes may not be worth it. I guess I'd see how far you could get while preserving the current APIs.

samselikoff:
I think you can go for it! I would definitely change some of the APIs to be both simpler and thus more easy to type but I know there's a lot of Mirage code out there so the backwards-incompat changes may not be worth it. I guess I'd see how far you could get while preserving the current APIs.

PR #1074, opened 2023-01-05
Establish precedence for embed over serializeIds

### LINKED ISSUE #1073: 0.1.46 breaks embedded records without `serializeIds: false`
We were unable to upgrade from 0.1.45 to 0.1.46 because the response output switched from arrays of embedded records to arrays of ids in the new version. We also saw relationships appear in the output that weren't there/not included previously.

`GET http://localhost:4200/ghost/api/admin/users/me/?include=roles`

0.1.45 output:
```
{
    users: [{
        ...
        roles: [{id: '1', ...}]
    }]
}
```

0.1.46 output:
```
{
    users: [{
        ...
        roles: ['1'],
        posts: ['1', '2', '3']
    }]
}
```

I haven't had a chance to dig into it properly yet but from a cursory reading of the changes in the release I think it's likely due to the removal of the [early return for embedded models here](https://github.com/miragejs/miragejs/pull/850/files#diff-8e3d45dd0e9ec504195499d7fafe3efc08756c349fce602ef6538c593aa563d8L719-R718). Rather than immediately returning the attrs when an embedded model is serialized, it's now passed to `_maybeAddAssociationIds` which then overwrites the embedded records - that was confirmed by adding `serializeIds: false` to our application serializer which bypassed the `_maybeAddAssociationIds` logic and kept the embedded records in the output.

NoneOfMaster:
#850 introduced support for selectively serializing embedded relationships. 

This PR should close #1073 which points out a collision which was also introduced between `serializeIds` (set to `always` as it is in [the `ActiveModelSerializer`](https://github.com/miragejs/miragejs/blob/master/lib/serializers/active-model-serializer.js#L5) and [the `RestSerializer`](https://github.com/miragejs/miragejs/blob/master/lib/serializers/rest-serializer.js#L5)) and the intended new functionality.

### DIFF
--- a/__tests__/external/shared/serializers/base/associations/embedded-collection-test.js
+++ b/__tests__/external/shared/serializers/base/associations/embedded-collection-test.js
@@ -170,4 +170,83 @@ describe("External | Shared | Serializers | Base | Associations | Embedded Colle
       ],
     });
   });
+
+  describe(`'embed' takes precedence over 'serializeIds'`, () => {
+    test(`it entirely overrides serializeIds`, () => {
+      const EmbedSerializer = Serializer.extend({
+        embed: true,
+        serializeIds: "always",
+      });
+
+      server.config({
+        serializers: {
+          application: EmbedSerializer,
+          blogPost: EmbedSerializer.extend({
+            include: ["author", "comments"],
+          }),
+        },
+      });
+
+      const post = server.schema.blogPosts.first();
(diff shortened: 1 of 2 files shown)


IanVS:
Thanks for the fix and the tests!

kevinansfield:
Thank you for the quick turnaround on the fix!

Issue #1079, opened 2023-01-23
`timing` setting does not respect `environment: 'test'` without setting `testConfig`


rmjohnson-olo:
I was surprised to find out my mirage calls in my tests were taking over 400ms, even though I had my environment set to `test`. The `timing` [documentation](https://miragejs.com/api/classes/server/#timing) led me to believe setting the environment was the only thing necessary to get the 0ms delay. After some debugging I realized [this](https://github.com/miragejs/miragejs/blob/d26b777ee9bcd6ba6f4e89c9df6c509e54476c2b/lib/server.js#L288-L290) only gets called if I pass in a `testConfig` function. So adding in `testConfig: () => {}` fixed the problem. However, this feels unexpected and should probably either be changed or the documentation should be updated to make that more clear.

As well, I had expected an explicit setting of `timing: 0` would also work. Unfortunately because `0` is falsey, [this line](https://github.com/miragejs/miragejs/blob/4df48d6f6b961a7a640b275154e1b1833c764dd5/lib/mock-server/pretender-config.js#L222) ignores `0` and defaults to the `400`. I'd expect `0` to be a valid value for `timing` even if my environment wasn't `test`.

Let me know if I can help with any of this, more than happy to explain further or propose a PR.

cah-brian-gantzler:
Please submit a PR. Looking at what you posted I would think that [this line](https://github.com/miragejs/miragejs/blob/4df48d6f6b961a7a640b275154e1b1833c764dd5/lib/mock-server/pretender-config.js#L222) would be where you would just fix it to understand 0.

You could just set the timing to 1 MS, would be about the same, update the docs accordingly

rmjohnson-olo:
@cah-brian-gantzler, I've opened up the following PR to address both of these issues: https://github.com/miragejs/miragejs/pull/1080

PR #1080, opened 2023-01-24
Don't require testConfig for timing to respect environment.

rmjohnson-olo:
Closes https://github.com/miragejs/miragejs/issues/1079

### DIFF
--- a/__tests__/internal/unit/server-test.js
+++ b/__tests__/internal/unit/server-test.js
@@ -75,7 +75,15 @@ describe("Unit | createServer", function () {
   });
 
   test("forces timing to be 0 in test environment", () => {
-    let server = new Server({ environment: "test" });
+    let server = createServer({ environment: "test" });
+
+    expect(server.timing).toEqual(0);
+
+    server.shutdown();
+  });
+
+  test("allows setting the timing to 0", () => {
+    let server = createServer({ timing: 0 });
 
     expect(server.timing).toEqual(0);
 
--- a/__tests__/internal/unit/server-test.js
+++ b/__tests__/internal/unit/server-test.js
@@ -73,6 +73,14 @@ describe("Unit | createServer", function () {
 
     server.shutdown();
   });
+
+  test("forces timing to be 0 in test environment", () => {
+    let server = new Server({ environment: "test" });
+
+    expect(server.timing).toEqual(0);
+
+    server.shutdown();
+  });
 });
 
 describe("Unit | Server #loadConfig", function () {
--- a/lib/server.js
+++ b/lib/server.js
@@ -285,9 +285,8 @@ export default class Server {
     }
 
     if (this.isTest()) {
-      if (config.testConfig) {
-        this.loadConfig(config.testConfig);
-      }
+      const testConfig = config.testConfig ?? function () {};
+      this.loadConfig(testConfig);
 
       if (typeof window !== "undefined") {
         window.server = this; // TODO: Better way to inject server into test env
--- a/lib/server.js
+++ b/lib/server.js
@@ -285,8 +285,7 @@ export default class Server {
     }
 
     if (this.isTest()) {
-      const testConfig = config.testConfig ?? function () {};
-      this.loadConfig(testConfig);
+      this.loadConfig(config.testConfig);
 
       if (typeof window !== "undefined") {
         window.server = this; // TODO: Better way to inject server into test env
@@ -340,7 +339,7 @@ export default class Server {
    * @hide
    */
   loadConfig(config) {
-    config.call(this);
+    config?.call(this);
     this.timing = this.isTest() ? 0 : this.timing || 0;
   }
 
--- a/__tests__/internal/move-after-handle-request/server-config-test.js
+++ b/__tests__/internal/move-after-handle-request/server-config-test.js
@@ -226,6 +226,22 @@ describe("Integration | Server Config", () => {
     expect(serializerMap.post).toEqual(RestSerializer);
   });
 
+  test("redefining timing options twice using the config method works", async () => {
+    expect.assertions(2);
+
+    server.config({
+      timing: 1000,
+    });
+
+    expect(server.timing).toEqual(1000);
+
+    server.config({
+      timing: 500,
+    });
+
+    expect(server.timing).toEqual(500);
+  });
+
   test("changing the environment of the server throws an error", () => {
     expect(() => {
       server.config({
--- a/lib/mock-server/pretender-config.js
+++ b/lib/mock-server/pretender-config.js
@@ -219,7 +219,7 @@ export default class PretenderConfig {
      @type Number
      @public
      */
-    this.timing = this.timing || config.timing || 400;
+    this.timing = config.timing ?? this.timing ?? 400;
 
     /**
      Sets a string to prefix all route handler URLs with.


cah-brian-gantzler (dismissed):

rmjohnson-olo:
@cah-brian-gantzler I took a look at the test failure this morning and realized the change from `||` to `??` broke another test. This was because it was initially setting the timing to `0` but then calling `config` to set the timing to another value. Previously with the `||`, the `0` on `this.timing` here:
https://github.com/miragejs/miragejs/blob/d26b777ee9bcd6ba6f4e89c9df6c509e54476c2b/lib/mock-server/pretender-config.js#L222

gets thrown out because it is falsey. However, when changed to `??`, it stays `0` because `this.timing` takes precedence. 

Changing [the order](https://github.com/miragejs/miragejs/pull/1080/commits/d6d223b913738838dd3f5a0af678c09552524454) such that the `config.timing` takes precedence fixes the tests, but I'm not confident that is the desired behavior and was hoping you might be able to shed some light there. Thanks!

cah-brian-gantzler:
That would not be the desired behaviour. If you configured a time for all the tests, then tried to override one test, it would never take effect and instead use the config.timing.

cah-brian-gantzler:
Is the test in question the one in the PR? or you referring to another test, if so can you link to that test so I can take a look.

rmjohnson-olo:
@cah-brian-gantzler, this existing [test](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/__tests__/internal/move-after-handle-request/server-config-test.js#L196-L228). It was failing on [this assertion](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/__tests__/internal/move-after-handle-request/server-config-test.js#L216).

cah-brian-gantzler:
For that test, `this.timing` should be undefined as [Here](https://github.com/miragejs/miragejs/pull/1080/files#diff-bdab195ccc6526e32f9cd04903927744666e61ff34e0bd9c93063affaa26dcc7L115). It was not assigned a default value. I am not finding where `this.timing` is assigned anything, so the ?? should still work. you are saying that `this.timing` was 0 when it got to the line with ??. Where was it set to 0 from the initial value of undefined?

rmjohnson-olo:
@cah-brian-gantzler it looked like `timing` gets set to `0` [here](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/__tests__/internal/move-after-handle-request/server-config-test.js#L17) in the `beforeEach`.

cah-brian-gantzler:
Now the question is why. It was already undefined. Setting it to zero wasnt needed.  There is also [this code](https://github.com/miragejs/miragejs/blob/master/lib/server.js#L345) pointed out earlier that makes tests run faster, but they set to 0, not undefined. 

Given all we see, while we could change all the internal code to treat 0 and undefined separately, there could be a lot of peoples tests that are doing what the code is doing now, setting timing to 0 to treat as undefined. Making this PR a breaking change. 

My suggestion would be to NOT do this PR, then update the docs to be more clear that undefined and 0 are the same thing for timing. Setting timing to 1 would pretty much accomplish the same goal you wanted by setting it to 0.

rmjohnson-olo:
>Now the question is why. It was already undefined. **Setting it to zero wasnt needed.** There is also [this code](https://github.com/miragejs/miragejs/blob/master/lib/server.js#L345) pointed out earlier that makes tests run faster, but they set to 0, not undefined.

It was needed for exactly the reason that I've created this PR, because it will default to 400 (and slow down your tests) otherwise. The [referenced code](https://github.com/miragejs/miragejs/blob/d26b777ee9bcd6ba6f4e89c9df6c509e54476c2b/lib/server.js#L345) only gets called if you pass in [config.testConfig](https://github.com/miragejs/miragejs/blob/d26b777ee9bcd6ba6f4e89c9df6c509e54476c2b/lib/server.js#L288) as I pointed out in my original issue.

olo-ntaylor:
> As a user I definitely would not think that 0 is the same as undefined. It means if I set my timing to 0 (meaning no delay), it will actually be 400? That feels like a bug, and explaining it in docs doesn't seem like the right solution.

Just want to pop in uninvited to 100000% agree with this assessment.

lib/server.js
```diff
@@ -285,9 +285,8 @@ export default class Server {
     }
 
     if (this.isTest()) {
-      if (config.testConfig) {
-        this.loadConfig(config.testConfig);
-      }
+      const testConfig = config.testConfig ?? function () {};
+      this.loadConfig(testConfig);
```
IanVS, line 289: I think I'm missing something, why do we need to call `loadConfig()` with a no-op if testConfig is undefined?  I get that it's one of the issues you identified originally, but it seems weird to call it with a no-op vs the config just being an optional parameter and a falsey value being handled in the `loadConfig` method itself.

cah-brian-gantzler:
It seems like the consensus is breaking change. Im in.

lib/server.js
```diff
@@ -285,9 +285,8 @@ export default class Server {
     }
 
     if (this.isTest()) {
-      if (config.testConfig) {
```
cah-brian-gantzler, line 288: All this code did was conditionally load the config. How does this effect timing?

cah-brian-gantzler (changes requested):

IanVS (commented):
I'll be honest I'm a bit confused how this is supposed to work.  The only real docs I found about testConfig is https://github.com/miragejs/miragejs/commit/8c5993732833948bc0194bc4242e6712f8d3d45a.  Is this something that's only applicable to ember-cli?  I use mirage without ember, and I don't have a "config file" per se, but rather I call `createServer` manually.  

Therefore, is this a bug that only effects ember users?  

I'm alright with taking this in two stages, getting this fix out first and then maybe queueing the other up for a major release.  But I'd like to understand the failure modes a bit better.  Is anyone able to whip up a quick reproduction of the issue where setting timing to 0 is the same as undefined?

lib/server.js
IanVS: I was confused by this too, but `loadConfig` is where the timing gets set, for some reason.

IanVS: I wonder if https://github.com/miragejs/miragejs/blob/d26b777ee9bcd6ba6f4e89c9df6c509e54476c2b/lib/server.js#L345 should be moved outside of `loadConfig` and into this `if (this.isTest())` conditional.

__tests__/internal/unit/server-test.js
```diff
@@ -73,6 +73,14 @@ describe("Unit | createServer", function () {
 
     server.shutdown();
   });
+
+  test("forces timing to be 0 in test environment", () => {
+    let server = new Server({ environment: "test" });
+
+    expect(server.timing).toEqual(0);
```
IanVS, line 80: Just to confirm, this test fails without your changes in this PR?

cah-brian-gantzler:
Yes I believe that testConfig is an ember-cli left over. I have been advocating for removing it altogether. There are other better ways to do it now. If the timing is getting set to 0 in the testConfig defined in the test, then it is a test only problem, remove the timing being set to 0 in the test. (ie update the test to reflect the new functionality where 0 and undefined do not equal)

lib/server.js
cah-brian-gantzler: Ahh, I believe this line should be 
```
this.timing = this.isTest() ? undefined : this.timing; 
```

__tests__/internal/unit/server-test.js
cah-brian-gantzler: With the proposed breaking change, this should expect undefined.

lib/server.js
IanVS: It's easy to get turned around in here.  😅

__tests__/internal/unit/server-test.js
rmjohnson-olo: Correct :+1:

rmjohnson-olo:
>I use mirage without ember, and I don't have a "config file" per se, but rather I call createServer manually.

I'm also using `createServer` manually. [createServer](https://github.com/miragejs/miragejs/blob/1652e006f54d67c01725a3b65c7d6bfcb662db89/lib/server.js#L43) just calls the server constructor, where [options that get passed in get set to the config](https://github.com/miragejs/miragejs/blob/1652e006f54d67c01725a3b65c7d6bfcb662db89/lib/server.js#L78) so this doesn't actually have anything to do with the "config file" as far as I can tell.

lib/server.js
rmjohnson-olo: >the config just being an optional parameter and a falsey value being handled in the loadConfig method itself.

I'm fine with this change too. I'll update :+1:

rmjohnson-olo: https://github.com/miragejs/miragejs/pull/1080/commits/ce991fd93f8c28c75ac234d92a46ee0c02a3f849

rmjohnson-olo:
Hey @cah-brian-gantzler @IanVS, hope you guys had a good weekend. =) Just wanted to check in to see how you guys want to proceed here. I'm hoping with the two tests we're on the same page now about what the problem is. I won't be offended if you guys want to close this PR and address them separately, I just wanted to do my OSS due diligence when I discovered these problems. =)

IanVS:
Thanks @rmjohnson-olo, I'm very glad you did speak up.  And thanks for showing the failing test, since I was obviously too lazy to just pull it down and try it myself.  😳 

@cah-brian-gantzler given this failing test, do you believe that the previous fix (using nullish coalescing to avoid treating 0 as falsey) is warranted here?  If so, do you still think it would be a breaking change, or simply a bugfix?

rmjohnson-olo:
@IanVS @cah-brian-gantzler just wanted to check in here to see if there was any interest in moving forward with this PR. Again, I've got a workaround working for us so I won't be offended if we decide to drop this for the time being.

cah-brian-gantzler:
@IanVS I believe the nullish coalescing change is required to make 0 not be treated as falsy. 

It is possible to be a breaking change is anyone is setting timing to 0 (instead of undefined) expecting it to then fallback to the next one, it would not any longer. Which is evidenced by the tests failing as they expected that behaviour. 

I do agree the change should be made, it is reasonable to expect setting it to 0 would honor 0. That being the case one could argue that this is a bug, and anyone using it it using an unintended feature :)

It looks like all the tests failed with that pesky second pretender instance that sometimes happens.

cah-brian-gantzler:
Didnt this start out as attempting to make 0 mean 0? But the end result of this code now is just changing an if check to a null coalesce check and moving it. 

It does the same thing so Im fine with it. Not sure it was actually needed though. 

For the original request of making 0 mean 0 and not false fallback to next timing. I would agree that should be done. I would also consider that a bug as it shouldnt have worked that way to begin with (even though the tests tested for it to work that way ☹️  ). It would be a breaking change if someone relied on it, but I would say they relied on an unintended behaviour.

rmjohnson-olo:
>But the end result of this code now 

Right, I reverted my code and added the failing unit test to illustrate the problem to @IanVS :

>Is anyone able to whip up a quick reproduction of the issue where setting timing to 0 is the same as undefined?


I'm happy to add that change back in if we're happy with it.

rmjohnson-olo:
>It does the same thing so Im fine with it. Not sure it was actually needed though.

To clarify, it doesn't do the same thing. With that change we're now calling `loadConfig` even if `testConfig` is not defined and therefore ` this.timing = this.isTest() ? 0 : this.timing || 0;` gets set.

IanVS:
>I'm happy to add that change back in if we're happy with it.

I'd be on board with that, since it sounds like we're all on the same page now that it was a bug.

rmjohnson-olo:
>I'd be on board with that, since it sounds like we're all on the same page now that it was a bug.

Awesome, so this still leaves the problem of [this assertion](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/__tests__/internal/move-after-handle-request/server-config-test.js#L216).

With just changing 

```
this.timing = this.timing || config.timing || 400;
```
to:
```
this.timing = this.timing ?? config.timing ?? 400;
```
That assertion will fail (because it depends on [this zero](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/__tests__/internal/move-after-handle-request/server-config-test.js#L17) being interpreted as falsey and therefore getting overriden by `config.timing`). How do we want to address that? Should `this.timing` or `config.timing` take precedence?

IanVS:
I'm not sure I'm tracking 100%, but it seems like calling `server.config()` with a new `timing` should work.  If it doesn't, that seems like a bug...  right?  Or am I misunderstanding?

rmjohnson-olo:
>I'm not sure I'm tracking 100%, but it seems like calling server.config() with a new timing should work. If it doesn't, that seems like a bug... right? Or am I misunderstanding?

I'd think so as well. I've added another test to hopefully illustrate the problem the current implementation: https://github.com/miragejs/miragejs/pull/1080/commits/f7e30b2117be0c5d8d23319399c96cfa30f39b6b

The reason the existing test works and this one doesn't is because `timing` is initially set to `0` (falsey) and therefore can be overriden with `server.config`. However, once timing gets set to `1000`, in my first `server.config` call, it can't get overridden with the `500` because `1000` is not falsey like `0` is.

IanVS (approved):
This looks good to me, thanks for your help and persistence, @rmjohnson-olo.

@cah-brian-gantzler what do you think?

cah-brian-gantzler (approved):
I will have to make these same changes in the mirage-pretender library. Sam had always wanted to extract pretender from MirageJS. We are setup for that. Not sure when we are actually going to do it. 

I would prefer if mirage-pretender and mirage-mws were brought under the mirageJS org.

IanVS:
>I would prefer if mirage-pretender and mirage-mws were brought under the mirageJS org.

Makes sense to me, but I think @samselikoff will need to do that, I don't have org privileges.

rmjohnson-olo:
Hey @IanVS @cah-brian-gantzler just wanted to check in here to make sure there wasn't anything you guys were waiting for me on. Thanks!

IanVS:
Thanks @rmjohnson-olo, we just need to coordinate how to release this.  I don't think there's anything we need from you at this point.  Appreciate the work here, and for keeping on top of us.  :)

IanVS:
I think we can merge this and deal with the extracted versions next.

rmjohnson-olo:
Thanks @IanVS, appreciate it!

PR #1089, opened 2023-08-31
add Middleware support via `server.middleware = [/*...*/]`

osi-jehrlich:
Relating to [Add Middleware support to Mirage #41 ](https://github.com/miragejs/discuss/issues/41).

# Middleware

This PR adds middleware to miragejs via a new `server.midleware = [...]`

Example middleware which randomly returns a 500 response:
```js
  function random500() {
    return (schema, req, next) => {
      return (Math.random() > 0.7)
        ? new Response(500, {}, 'no')
        : next();
    }
  }
```
  
Routes which use the middleware defined above: 
_(Edited to match latest update)_
```js
  routes() {
    this.middleware = [
      random500(),
      addABunchOfResponseHeaders(),
      // more middleware can be added here...
    ]

    // Subsequent routes will have that middleware
    this.get('/users', (schema, req) => {
      return new Response(204, {}, null);
    });

    this.middleware = [
      // differentMiddleware
    ]
     
    this.get('/frogs', (schema, req) => {
      return new Response(204, {}, null);
    });
  }
```
# Not included

1. There's a bit of funkyness when middleware wants to modify the response from a route handler because the route handler might return a `Response` object or it might return something else altogether.  I imagine this could be addressed separately.

2. `async` middleware - again, could be added after (if this is the approach this repo heads in).

### DIFF
--- a/lib/route-handler.js
+++ b/lib/route-handler.js
@@ -97,9 +97,7 @@ export default class RouteHandler {
 
       result = this.handleWithMiddleware(request, [
         ...middleware,
-        () => {
-          return this.handler.handle(request);
-        },
+        (_, req) => this.handler.handle(req),
       ]);
     } catch (e) {
       if (e instanceof MirageError) {
@@ -123,10 +121,10 @@ export default class RouteHandler {
     return this._toMirageResponse(result);
   }
 
-  handleWithMiddleware(request, remainingMiddleware) {
-    const current = remainingMiddleware[0];
-    return current.call(null, this.schema, request, () => {
-      return this.handleWithMiddleware(request, remainingMiddleware.slice(1));
+  handleWithMiddleware(request, middleware) {
+    const [current, ...remaining] = middleware;
+    return current(this.schema, request, (req = request) => {
+      return this.handleWithMiddleware(req, remaining);
     });
   }
 
--- a/lib/route-handler.js
+++ b/lib/route-handler.js
@@ -17,9 +17,17 @@ function createHandler({
   path,
   rawHandler,
   options,
+  middleware,
 }) {
   let handler;
-  let args = [schema, serializerOrRegistry, rawHandler, path, options];
+  let args = [
+    schema,
+    serializerOrRegistry,
+    rawHandler,
+    path,
+    options,
+    middleware,
+  ];
   let type = typeof rawHandler;
 
   if (type === "function") {
@@ -52,10 +60,12 @@ export default class RouteHandler {
     options,
     path,
     serializerOrRegistry,
+    middleware,
   }) {
     this.verb = verb;
     this.customizedCode = customizedCode;
     this.serializerOrRegistry = serializerOrRegistry;
+    this.middleware = middleware || [];
     this.handler = createHandler({
       verb,
       schema,
@@ -67,14 +77,14 @@ export default class RouteHandler {
   }
 
   handle(request) {
-    return this._getMirageResponseForRequest(request)
+    return this._getMirageResponseForRequest(request, this.middleware)
       .then((mirageResponse) => this.serialize(mirageResponse, request))
       .then((serializedMirageResponse) => {
         return serializedMirageResponse.toRackResponse();
       });
   }
 
-  _getMirageResponseForRequest(request) {
+  _getMirageResponseForRequest(request, middleware = []) {
     let result;
     try {
       /*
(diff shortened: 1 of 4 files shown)


osi-jehrlich:
Just realized an even simpler API was staring me right in the face, similar to the API to set a namespace:

```ts
  routes() {
    this.middleware = [
      random500(),
      addABunchOfResponseHeaders(),
      // more middleware can be added here...
    ]

    // Subsequent routes will have that middleware
    this.get('/users', (schema, req) => {
      return new Response(204, {}, null);
    });

    this.middleware = [
      // differentMiddleware
    ]
     
    this.get('/frogs', (schema, req) => {
      return new Response(204, {}, null);
    });
  }
```

I've update this PR to have this API.  It still has `withMiddleware`, too, but that's now optional.  Thoughts?

lib/route-handler.js
```diff
@@ -85,7 +95,12 @@ export default class RouteHandler {
         this.handler.setRequest(request);
       }
 
-      result = this.handler.handle(request);
+      result = this.handleWithMiddleware(request, [
+        ...middleware,
+        () => {
+          return this.handler.handle(request);
```
IanVS, line 101: With this approach, if middleware wants to change the request, it needs to mutate the `request` parameter directly before calling `next()`, right?  What would you think about also allowing middleware to call `next(request)`, to be able to pass a cloned or even completely new request object into the subsequent handler?  I think something like this could work, while still allowing `next()` to be called without arguments, right?

```suggestion
        (req = request) => {
          return this.handler.handle(req);
```

IanVS (commented):
Hi @osi-jehrlich, thanks for the contribution!  I have to admit that after the original maintainer of this project left, we're just a skeleton crew keeping the lights on, so I'm a bit reluctant to add new features.  That said, this does seem like it could be pretty useful, and you've added both tests and types so I feel a bit more comfortable with it.  I'd argue against adding two different APIs that accomplish the same thing, and your new API feels a bit more similar to how other parts of mirage work, so I'd suggest removing `withMiddleware`.  

I'd also like to hear what @cah-brian-gantzler thinks of this approach.

lib/route-handler.js
osi-jehrlich: I like the idea quite a bit; it'll need a couple more lines of code and type definitions, too - I'll get on that.  I haven't made new `Requests` from old `Requests`, but as long as all the data is available to copy over it sounds sound.

Removing `withMiddleware` seems like a good idea, too.  I think it's easy enough for library consumers to write a standalone version of that if they need to once the simpler API is available.  We can always add it at some point if there's demand.

osi-jehrlich: Updated :-)

IanVS:
Thanks, I'd like to hear what @bgantzler thinks of this approach, but he's been busy lately.

cah-brian-gantzler:
Given what I see Im fine with it. 

Future changes could be that the default handler function is supplied as the last element. While I have already extracted the default handlers so that we could supply a Pretender or the MSW variation. With this change Im wondering if there isnt a way to further extract that to make the Pretender and MWS version BE just another middleware (I realize they are since the handler is in the last spot, but I bet theres room for improvement and simplification of the abstract handler). Just a thought, could be wrong. Would have to explore to see if that could be done or not. 

I am fine with this PR. I understand it, and its a pretty simple change considering what its doing.

osi-jehrlich:
FYI, I've created a PR with documentation to the repo for the main website.  If you get a chance to take a look I'd appreciate any and all feedback.

Is there anything else I can do (or need to do) to help move this PR toward merge/release?

dodalovicgran:
Hi, how to call async methods in middlewares?

dodalovicgran:
Hi, how to call async methods in middlewares?

osi-jehrlich:
1. We didn't add explicit async support into middleware yet.
2. That said, I believe some asynchronous middleware should already work, see [these tests](https://github.com/osi-jehrlich/miragejs/blob/async-middleware-tests/__tests__/internal/move-after-handle-request/middleware-async-test.js
) (in a branch where the only change is adding those tests).  Maybe it only works if it's the first middleware or something like that...

Perhaps we can move this discussion to the miragejs/discuss repo?

dodalovicgran:
Thank you @osi-jehrlich This worked for me!

PR #1103, opened 2024-04-18
Fix stale request object when an explicit serializer type is passed

### LINKED ISSUE #885: FunctionRouteHandler.serialize() uses stale query params when passed an explicit serializer type
The song goes a little like this.

- [SerializerRegistry sets a `this.request` prop on itself when you call `serialize()` on it.](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/lib/serializer-registry.js#L23)
- [But/and, when you call `serializerFor()` on it, it uses that saved `this.request` state to instantiate the serializer.](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/lib/serializer-registry.js#L70)
- So if you ever call `serializerFor()` yourself, without passing through `serialize()`, you get stale state in your serializer instance, including queryParams and the like. 😑 (Specifically, you get data from the last thing that was passed through `SerializerRegistry.serialize()`.)
- Oh hey, [`FunctionRouteHandler.serialize()` calls `SerializerRegistry.serializerFor()` if you pass it an explicit serializer type.](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/lib/route-handlers/function.js#L29)
- Then it almost recovers, by [passing the _proper_ request to `serializer.serialize()`](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/lib/route-handlers/function.js#L36)... 
- Then Serializer [throws it away.](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/lib/serializer.js#L364) Well, it was a nice try.
- Later, JSONAPISerializer uses the `this.request` prop that was populated when it was instantiated to answer [things like `getQueryParamIncludes()`.](https://github.com/miragejs/miragejs/blob/b965b8a058523dac0d6ab4792826887154e25368/lib/serializers/json-api-serializer.js#L560)
- And by that point it's probably trying to include an association that doesn't even exist on the model for the records you passed it, and you get an explosion.

zeppelin:
Fixes #885. (Was likely [introduced here](https://github.com/miragejs/miragejs/commit/e7e041f765e04f8ea79d1626dc7ed140a2d52945#diff-f1b016512c8f528ede71d29cc11835a4a81f451bc0bf53abdf6fc715d8b3e65aR38-R40))

Although it does solve the issue in the least invasive way, the bug could persist in setups where the `serialize()`  method is overridden, because while the `request` argument itself is documented, the [doc suggests](https://miragejs.com/api/classes/serializer/#serialize) it's fine to throw it away. Also, it doesn't quite fix the underlying issue, namely that serializer & serializer-registry state isn't reset across requests. For example, when `serializerRegistry.serializerFor` is used to obtain the serializer, the request object must be assigned manually to the serializer, otherwise it'd break the same as before.

Nevertheless, all other solutions would end up being _much_ more complex and have more likelihood of breaking some Mirage setups.


zeppelin:
Ping @IanVS :)

nickschot:
We also ran into this today, landed at the same fix.

PR #1104, opened 2024-06-12
Breaking: Make createServer async to support MSW.

IanVS:
This is to support MSW startup, which is async.  It has the downside of also making startup using pretender also async, even though it doesn't need to be.  Perhaps there's a bit of an inversion we can do here, so that the interceptors import from miragejs core, and users interact only directly with the interceptors.  I'll try experimenting around with something like that, but in the meantime, we could release this as an alpha version in case anyone wants to experiment with mirage-msw.  To that end, this PR is targeting a new `next` branch, to keep the main branch clean in case we need to release something (unlikely, I know).

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -39,8 +39,10 @@ const defaultInflector = { singularize, pluralize };
  * @param {Object} options.factories Server factories
  * @param {Object} options.pretender Pretender instance
  */
-export function createServer(options) {
-  return new Server(options);
+export async function createServer(options) {
+  const server = new Server();
+  await server.config(options);
+  return server;
 }
 
 /**
@@ -75,7 +77,6 @@ export default class Server {
    */
   constructor(options = {}) {
     this._container = new Container();
-    this.config(options);
 
     /**
       Returns the Mirage Db instance.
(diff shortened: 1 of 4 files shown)


lib/server.js
```diff
@@ -39,8 +39,10 @@ const defaultInflector = { singularize, pluralize };
  * @param {Object} options.factories Server factories
  * @param {Object} options.pretender Pretender instance
  */
-export function createServer(options) {
-  return new Server(options);
+export async function createServer(options) {
+  const server = new Server();
+  await server.config(options);
+  return server;
```
cah-brian-gantzler, line 45: Taking the options away from the constructor means you can create a server that isnt configured. Looking at the code, the only thing you need awaited if the interceptor start, which is the very last line of config. What about keeping config non async and instead doing 

```
export async function createServer(options) {
  const server = new Server(options);
  await server.start();
  return server;
```
Move the code from config to its own method, start
```
async start() {
     await this.interceptor.start?.();
}
```

cah-brian-gantzler (changes requested):

lib/server.js
IanVS: Great suggestion, thanks.

cah-brian-gantzler (approved):

PR #1107, opened 2024-06-14
Use correct ESM/CJS extensions

IanVS:
This fixes our esm output file to use the correct `.mjs` extension and modernizes our package.json a bit, to use `"exports"` (supported since Node 12.7) rather than the non-standard `main/module/types` fields.  

It also adds `"type": "module"` to our package.json, since that's how the files are written (esm in .js files).  I needed to make a few adjustments to get jest working properly as well.

I also removed the very-old `engines` field from package.json (I think we can do without it), and added a `.node-version` file so that tools like [fnm](https://github.com/Schniz/fnm) will pick it up and use the reasonable node version while we're developing.


cah-brian-gantzler:
Sorry but not up to date on this level of things. I wouldnt know if you were doing it right or not. I would not be able to review this.

PR #1110, opened 2024-06-17
Avoid stringifying valid Response body types

IanVS:
This only works with mirage-msw, because pretender _always_ tries to deal with the response body as a string or `ArrayBuffer`.  But, together with one small change in mirage-msw (PR incoming soon), we can allow streaming mock responses when using MSW.  

I needed to use `jest-fixed-jsdom`, because otherwise jest doesn't know what to do with `ReadableStream` even though it's been native to web and node since at least Node 18.  

I also tried to write an explicit test for ReadableStream using mirage-msw, but it's not possible right now, because MSW throws an error if you try to use the browser worker in node.js.  I may explore switching to vitest, to use its new browser mode, so we can run node tests in node, and browser tests in a real browser.  But that will need to come later.


lib/server.js
```diff
@@ -719,7 +719,7 @@ export default class Server {
   }
 
   _serialize(body) {
-    if (typeof body === "string") {
+    if (typeof body === "string" || body instanceof ReadableStream) {
       return body;
     } else {
       return JSON.stringify(body);
```
cah-brian-gantzler: This method use to always return a string (either way), So Why in the signature below was it `String | {}`? Isnt an instance of ReadableStream an `object` which is what `typeof {}` is. Should the description be `String | ReadableStream`?

In reality this is all about should we call JSON.stringify or not. Think this will happen again and we need to add another type to it? Should the if be replaced with a method call `shouldSerialize(body)` that could be overridden if needed?

cah-brian-gantzler (commented):
I am fine either way, but did have a question trying to understand it

lib/server.js
IanVS: Good question, that confused me at first as well.  It seems the `Request` itself can have data that is an object (should be typed as `object` not `{}`, technically).  But when it is run through the route handler, a `serialize` is normally called on it, which typically stringifies the `data` on the request before `.toRackResponse()` is called.  The types in this project are not terribly exact, since they are written separately from the code itself.  I'd love to rewrite mirage in typescript, so we could be sure the exported types are correct, but so much of this library is dynamic that I think I might have a rough time of it.  I'm tempted to try sometime, though.  Until then, I think the types have to just be a "best effort" affair.

>Should the if be replaced with a method call shouldSerialize(body) that could be overridden if needed?

That's an interesting thought...  Since there's a [well-defined](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response) set of possibilities for response bodies, I wonder if we should just be exhaustive here.  OR, we can leave it up to the interceptors to serialize if necessary, since it's really only MSW that supports types other than strings.  Then pretender can throw a helpful error message if it finds a `FormData` or `ReadableStream` or whatever.  Though, if pretender adds support for other types, it would be annoying if we also need to make updates.  For example, pretender does support `ArrayBuffer` since https://github.com/pretenderjs/pretender/pull/148, but I think our `_serialize` function will still prevent that from working.  

So, I think I've talked myself in a circle, and the best thing we can do is to only serialize POJOs, and all other valid types of Response bodies should be passed through as-is, both to pretender and to msw.  Then if pretender starts being maintained again, it could also add support for other types.

cah-brian-gantzler (approved):

PR #1116, opened 2024-10-04
fix: Use `logging` from config

TSenter:
Currently when setting up the config, we're looking at if config.logging is undefined - if it's not, we're using this.logging. This means that whatever value you pass in as config.logging will be ignored.

### DIFF
--- a/lib/server.js
+++ b/lib/server.js
@@ -250,7 +250,7 @@ export default class Server {
       @return {Boolean}
       @public
     */
-    this.logging = config.logging !== undefined ? config.logging : undefined;
+    this.logging = config.logging || undefined;
 
     this.testConfig = this.testConfig || undefined;
 
--- a/lib/server.js
+++ b/lib/server.js
@@ -250,7 +250,7 @@ export default class Server {
       @return {Boolean}
       @public
     */
-    this.logging = config.logging !== undefined ? this.logging : undefined;
+    this.logging = config.logging !== undefined ? config.logging : undefined;
 
     this.testConfig = this.testConfig || undefined;
 


IanVS:
Thanks for the PR, @TSenter.  Can you describe what issue this is solving for you?

cah-brian-gantzler:
I think looking at the code change it is pretty clear it is wrong. However given the fixed code, I would think you would write it the same format as the line below it for consistency

TSenter:
> Thanks for the PR, @TSenter. Can you describe what issue this is solving for you?

Sure! Currently when setting up the config, we're looking at if `config.logging` is undefined - if it's not, we're using `this.logging`. This means that whatever value you pass in as `config.logging` will be ignored.

TSenter:
> I think looking at the code change it is pretty clear it is wrong. However given the fixed code, I would think you would write it the same format as the line below it for consistency

Good point! I've updated this PR.

IanVS:
Thanks, I just like to have a description in PRs about the changes being made.

lib/server.js
```diff
@@ -250,7 +250,7 @@ export default class Server {
       @return {Boolean}
       @public
     */
-    this.logging = config.logging !== undefined ? this.logging : undefined;
+    this.logging = config.logging || undefined;
 
     this.testConfig = this.testConfig || undefined;
```
IanVS, line 255: I haven't looked at this code closely enough.  Why does this kind of line exist?  The line being changed above is different, it's checking for `config`, whereas the line here is checking `this`.

IanVS (changes requested):
Can you please provide a "how to test" section?  This can be either an automated test that you add, or a set of steps that can be followed showing the previous (broken) behavior without this change, and the desired behavior after the change.

lib/server.js
TSenter: I'm not sure about the line you commented on, that's nothing that I changed.

TSenter:
@IanVS, @cah-brian-gantzler - would you mind looking at this again? I know it's been a while but I still think this is a good fix to have.

lib/server.js
cah-brian-gantzler: I think 253 is wrong and the value after the ? should have been config.logging.   The both 253 and 255 says is the value is not undefined use else, else assign undefined, isnt that the same as assign undefined. Unless they are trying to turn null into undefined. 

While weird, I think line 253 does fix the problem for the original line 253

IanVS:
The change seems reasonable, but I asked for some instructions on how to test.

cah-brian-gantzler:
OK, now I understand better what is actually happening here, the line should be
```
this.logging = config.logging !== undefined ? config.logging : this.logging;
```
or 
```
this.logging = config.logging ?? this.logging;   // The new ?? operator makes that easier to write. 
```

The intent is, if the config.logging is set to a value (presumably true or false), use it, else use this.logging, which is the reason for the test !== undefined.   However || returns the first truthy value, so `false || undefined` returns undefined, not false. The new ?? operator makes that easier to write. 

logging equal to undefined means different things based on environment. Look at lines 218-220.
```
 Set to `true` or `false` to explicitly specify logging behavior.

      By default, server responses are logged in non-testing environments. Logging is disabled by default in testing, so as not to clutter CI test runner output.

      For example, to enable logging in tests, write the following:
```

Now for testings, line 226 shows turning logging on and off after the server is created. 
```js
      test('I can view all users', function() {
        server.logging = true;
        server.create('user');

        visit('/users');
        // ...
      });
```

The line of this change happens when the server is created (passed on the constructor so)

```
createServer({
  routes() {
    this.get("/api/users", () => [
      { id: "1", name: "Luke" },
      { id: "2", name: "Leia" },
      { id: "3", name: "Anakin" },
    ])
  },
  logging: false
})
```
Passing logging on the hash that is passed on the createServer invokes this code as the create calls the config method on line 79. However you could also call config on the server `server.config({ logging: false })` as well. If you are setting just the one value, `server.logging = false` would be simpler, but calling `server.config({...})` would allow you to change many values at once.

So you can test by passing on the create or by calling config. The key would be that this.logging is NOT changed unless there is a logging present on the hash, and the key there is the value false means present, use it, although it is a falsey value. 

The line change needs to make sure that if we pass false (we do not want logging) that it remains false and does not get converted to undefined.

Issue #1113, opened 2024-07-18
Cannot name Factory property as "modelName"


LiaoJimmy:
## Example Environment
https://stackblitz.com/edit/vitejs-vite-valihw?file=src%2FMirage.ts

## Scenario
- Create a scooter modal
- Create a scooter factory with property: `modelName`
- Create a scooter seed

### Expected Result
- Get scooter list

### Actual Result
Get mirage error
```
miragejs.js?v=9692364e:9061 Uncaught TypeError: Cannot read properties of undefined (reading 'find')
    at Child.isNew (miragejs.js?v=9692364e:9061:52)
    at Child.save (miragejs.js?v=9692364e:8976:14)
    at Schema.create (miragejs.js?v=9692364e:11021:34)
    at Object.create (miragejs.js?v=9692364e:10971:31)
    at Server.create (miragejs.js?v=9692364e:12031:34)
    at Server.createList (miragejs.js?v=9692364e:12094:29)
    at Object.seeds [as default] (Mirage.ts:22:14)
    at Server.config (miragejs.js?v=9692364e:11804:31)
    at new Server (miragejs.js?v=9692364e:11714:10)
    at startMirage (Mirage.ts:4:3)
 ```

## Note
If you rename `modelName` to `modelName123`, it's working correctly.

## Screenshot
![mirage](https://github.com/user-attachments/assets/8744bad0-3ee0-441f-8571-dc8f8bd0bffd)

AhmedKhaled590:
Hi, Can u please assign me to work on this issue?

IanVS:
@AhmedKhaled590 feel free to give it a shot!

PR #1117, opened 2024-10-11
fix: prevent using reserved words while defining models in miragejs

AhmedKhaled590:
Resolving: #1113 

This PR introduces a series of assertions to prevent potential conflicts when defining models in Mirage. Specifically, it ensures that developers do not define properties with names that are reserved for internal use by the Mirage framework. The following properties are reserved and now protected through assertions: modelName, _schema, fks, attrs, _tempAssociations, and __isSavingNewChildren.

The assertions check whether any of these reserved property names are being used in the model definition. If a reserved name is detected, an error message will be thrown, providing clear feedback to the developer about the conflict and recommending they choose a different property name.

Additionally, when creating a model, the code also validates against the attributes (attrs) provided to ensure none of their keys conflict with the reserved properties. This safeguard helps maintain the integrity of internal Mirage functionality and avoids unexpected behavior.

### DIFF
--- a/lib/orm/model.js
+++ b/lib/orm/model.js
@@ -65,29 +65,44 @@ class Model {
     assert(modelName, "A model requires a modelName");
 
     assert(
-      !this.modelName,
+      !!this.modelName,
       "The 'modelName' property is reserved by Mirage for internal use. Avoid using this property in your model to prevent conflicts."
     );
     assert(
-      !this._schema,
+      !!this._schema,
       "The '_schema' property is reserved by Mirage for internal use. Avoid using this property in your model to prevent conflicts."
     );
     assert(
-      !this.fks,
+      !!this.fks,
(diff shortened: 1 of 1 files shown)


LiaoJimmy:
Hi, @AhmedKhaled590 .
In mirage docs, no one mentions these words are reserved words.
I wonder that we can use these words as property name or not?
Or maybe we should also update the docs and prevent other developers use these words as property name?

AhmedKhaled590:
Hi @LiaoJimmy 
Based on the issue reported and my debugging, it appears that using these specific words as property names causes different parts of the code to break. Therefore, we can't use them. Additionally, we should update the documentation to reflect this.

AhmedKhaled590:
Hi! I'm not sure if any of the maintainers review this or not.
Thanks!

IanVS:
I will take a look when I get a chance, thanks for the PR.

PR #1118, opened 2025-03-12
fix: recreating a new model with same id after getting deleted is not possible due to id already used.

pomm0:
How to reproduce:

* `server.create('user', { id: 1 });`
* `server.schema.users.find(1).destroy();`
* `server.create('user', { id: 1 });`
===
`Attempting to use the ID 1, but it's already been used`

I left the nextId untouched, because it would only make sense to reset from, lets say, 10 to 9 if the last element gets removed, but not, if 9 gets removed but we already have 15 models. And to reset for this only case is not worth to implement and maintain, in my opinion.

### DIFF
--- a/__tests__/external/shared/orm/destroy-test.js
+++ b/__tests__/external/shared/orm/destroy-test.js
@@ -34,6 +34,20 @@ describe("External | Shared | ORM | destroy", () => {
     expect(server.db.users).toHaveLength(2);
   });
 
+  test("destroying a model marks id as unused and allows creating a new model with same id", () => {
+    expect(server.db.users).toHaveLength(3);
+
+    let link = server.schema.users.find(1);
+    link.destroy();
+
+    server.schema.users.create({ id: 1, name: "Ganon", evil: true });
+    const ganon = server.schema.users.find(1);
+
+    expect(ganon).not.toBeNull();
+    expect(ganon?.name).toBe("Ganon");
+    expect(server.db.users).toHaveLength(3);
+  });
+
   test("destroying a collection removes the associated records from the db", () => {
     expect(server.db.users).toHaveLength(3);
 
--- a/__tests__/internal/unit/db/identity-manager-test.js
+++ b/__tests__/internal/unit/db/identity-manager-test.js
@@ -82,4 +82,13 @@ describe("Unit | Db | IdentityManager", function () {
 
     expect(true).toBeTruthy();
   });
+
+  test(`unset removes id from used ids`, () => {
+    let manager = new IdentityManager();
+    manager.set("abc");
+    manager.unset("abc");
+    manager.set("abc");
+
+    expect(true).toBeTruthy();
+  });
 });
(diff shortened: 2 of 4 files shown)


IanVS:
Hmmmm, I'm not sure about this one, whether IDs should be reused.  I'm guessing this is an intentional design, and I'm not sure we should change the behavior.

Can you talk more about why you think this should be the behavior?

pomm0:
It may be more relevant for string ids than for integer ids. In a normal DB-scenario the id would not be reused again, I believe. We have synthetic string ids like `e43e326b-da7e-4768-93ff-58509f4b2501:d3717d82-90cf-4ee2-aba4-2a95522ed69c` where each part (splitted by `:`) is a id from different services and will always be the same. The api allows deleting and creating these records multiple times and right now it is not possible to mock that behaviour out of the box with mirage.

I think when mirage lets us set the ID on creation of models, and allows deleting them, it should also allow creating a new record with the same ID again.

If this is a design decision, then I would argue it should also not reset used-ids when a [`db.users.remove()`](https://github.com/miragejs/miragejs/blob/605b35976fa00852ec61a5c711bbd27dbdeaa507/lib/db-collection.js#L311) is performed. Feels like inconsistency to me.

IanVS:
Thanks for explaining.  @cah-brian-gantzler what do you think here?

cah-brian-gantzler:
I think the that [db.users.remove()](https://github.com/miragejs/miragejs/blob/605b35976fa00852ec61a5c711bbd27dbdeaa507/lib/db-collection.js#L311) is removing all records and since it was a complete removal, said why not just reset the identities as well.

BTW the code you are pointing to is in 0.2.0-alpha.3 and if you look at line 316, it calls an unset on the identity manager to do what you are suggesting (that line is missing from master). So I think the proposed fix was already suggested.

pomm0:
I'm happy to rebase and target master, if thats more likely to get released earlier. Whats the plan with version 0.2?

@cah-brian-gantzler I'm not sure I get you right, line 316 has been added by me with this PR. `unset` on the identity manager does not exist, I introduced it. The link I provided with my last comment was a link into my commit, with my changes.

I think mirage should support this, a quick test with a [postgres DB proved](https://www.db-fiddle.com/f/4jyoMCicNSZpjMt4jFYoz5/16434), that it is possible to insert a new model with the same ID (note the unique constraint) after it gets deleted. As far as I understand mirage and its schema/DB we are talking here, it should just behave like a real DB. Or is there something I'm missing?

cah-brian-gantzler:
Sorry, when I went to the link all I had was it was on branch 605b359, didnt realize that was your branch. 

I would agree that the unset should be there. It clearly goes checks for the a truthy value [here](https://github.com/miragejs/miragejs/blob/master/lib/identity-manager.js#L47). It makes sense to set it to false when it is removed. Dont know why it wasnt done that originally. 

I would approve of that.

